PR #20, opened 2023-10-12
[iOS] - Added EULA license file to pass release validation requirements.

GabrielMedAlv:
Added Non-Windows EULA license file to pass release requirements.


source/EULA-Windows.txt
```diff
@@ -0,0 +1,40 @@
+MICROSOFT SOFTWARE LICENSE TERMS
```
jwoo-msft, line 1: I don't think that you need EULA-Windows.text

source/EULA-Non-Windows.txt
```diff
@@ -0,0 +1,41 @@
+MICROSOFT SOFTWARE LICENSE TERMS
```
jwoo-msft, line 1: Do your team have a guideline on which license to include?
You may be fine, but it would be clear following the guideline since this project has no dependency on Windows.

jwoo-msft (approved):

source/EULA-Windows.txt
GabrielMedAlv: You are right Joseph. I checked podspec and only mentioned non-Windows file. I removed it. Thanks.

source/EULA-Non-Windows.txt
GabrielMedAlv: removed the Windows one as you mentioned not necessarily due to podspec requirements. I will check with the team the right licences to add.

ardlank (approved):

PR #35, opened 2023-10-31
"Requires" and "fallback" for Adaptive Cards at root level : Shared library land iOS side logic

GabrielMedAlv:
Added a requirement to have requires and fallback fields at root JSON level of adaptive cards. This feature will be useful for future cards that have specific requirements to be displayed at all.

**Quick list of changes**

**Shared Library**
- Moved "_parseFallback_" and "_parseRequires_" functions from "_BaseElement_" to "_ParseUtil_" class to be used at root level.
- Call moved parsed functions in parse card fallback logic inside "_SharedAdaptiveCard_" class.
- Added require and fallback components into "_AdaptiveCard_" class.
- Moved "_meetRequirements()_" functions to "_ParseUtil_" class.
- Added  parsing of "requires" into "_SharedAdaptiveCard_" class.

**iOS Specific**
- Add root fallback into recursive validation of current elements fallback in "_ACRView_" class.

**Test Json** 
(Also included as an example into the iOS visualizer)

`{
  "type": "AdaptiveCard",
  "version": "1.6",
  "requires": {
    "responsiveLayout": "1.0"
  },
  "fallback": {
    "type": "TextBlock",
    "text": "Adaptive Cards not supported"
  },
  "body": [
    {
      "type": "TextBlock",
      "text": "targetWidth = veryNarrow",
      "wrap": true,
      "targetWidth": "veryNarrow"
    },
    {
      "type": "TextBlock",
      "text": "targetWidth = narrow",
      "wrap": true,
      "targetWidth": "narrow"
    },
    {
      "type": "TextBlock",
      "text": "targetWidth = standard",
      "wrap": true,
      "targetWidth": "standard"
    },
    {
      "type": "TextBlock",
      "text": "targetWidth = wide",
      "wrap": true,
      "targetWidth": "wide"
    }
  ],
  "msTeams": {
    "width": "full"
  }
}
`

**Screenshot**

Simulator Screenshot - iPhone 14 - 2023-10-30 at 19 57 59


source/shared/cpp/ObjectModel/SharedAdaptiveCard.h
```diff
@@ -123,6 +141,14 @@ class AdaptiveCard
     std::vector<RemoteResourceInformation> GetResourceInformation();
 
     CardElementType GetElementType() const;
+    
+    std::unordered_map<std::string, AdaptiveCards::SemanticVersion>& GetRootRequires();
+    const std::unordered_map<std::string, AdaptiveCards::SemanticVersion>& GetRootRequires() const;
```
ardlank, line 146: @GabrielMedAlv Do we need both GetRootRequires() ?

GabrielMedAlv: I put both methods more to conform similar structure as other functions in the same class like:

`std::vector<std::shared_ptr<BaseCardElement>>& GetBody();
    const std::vector<std::shared_ptr<BaseCardElement>>& GetBody() const;
    std::vector<std::shared_ptr<BaseActionElement>>& GetActions();
    const std::vector<std::shared_ptr<BaseActionElement>>& GetActions() const;`

samples/v1.6/Tests/Fallback.Root.Recursive.json
```diff
@@ -0,0 +1,33 @@
...
+   "$schema":"http://adaptivecards.io/schemas/adaptive-card.json",
+   "version":"1.6",
+   "fallback":{
+      "type":"TextBlock",
+      "text":"Adaptive Cards not supported"
+   },
+   "body":[
+      {
+         "type":"TextBlock",
+         "text":"This TextBlock should always render",
```
ardlank, line 12: This should also not render. Only root fallback should render

GabrielMedAlv: Yes, You are right. And that works when the CanFallbackToAncestor is set to true. to throw the exception in:

if (bCanFallbackToAncestor && fallbackType != FallbackType::Drop)

Need to verify which components have this flag on true or false.

source/shared/cpp/ObjectModel/SharedAdaptiveCard.cpp
```diff
@@ -260,6 +347,32 @@ std::shared_ptr<ParseResult> AdaptiveCard::Deserialize(const Json::Value& json,
...
+
+bool AdaptiveCard::MeetsRootRequirements(std::unordered_map<std::string, AdaptiveCards::SemanticVersion> requiresSet)
+{
+    std::unordered_map<std::string, AdaptiveCards::SemanticVersion> featuresSupported = GetFeaturesSupported();
+    
+    for (const auto &featureToCheck : requiresSet)
+    {
+        auto foundFeature = featuresSupported.find(featureToCheck.first);
+        
+        if (foundFeature == featuresSupported.end())
```
ardlank, line 361: Check case sensitive here

ardlank (approved):

karthikgopal36 (approved):

PR #70, opened 2024-01-12
Ardy/fix responsive layout

ardlank:
Fix 2 Edge cases with Responsive layout and Fallback. 

1. When the there is an element that needs to have a root fallback and that element is deeply nested in the JSON.
2. Target width elements should not fallback to the root if their target element is not met.

### DIFF
--- a/samples/v1.5/Scenarios/NestedFallbackBody.json
+++ b/samples/v1.5/Scenarios/NestedFallbackBody.json
@@ -1,0 +1,88 @@
+{
+	"type":"AdaptiveCard",
+	"$schema":"http://adaptivecards.io/schemas/adaptive-card.json",
+	"version":"1.6",
+	"requires":{
+		"responsiveLayout":"1.0"
+	},
+	"fallback":{
+		"type":"TextBlock",
+		"text":"Should not render root fallback"
+	},
+	"body":[
+		{
+			"type":"Container",
+			"items":[
+				{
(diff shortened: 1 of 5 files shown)


GabrielMedAlv (approved):

source/android/adaptivecards/src/main/java/io/adaptivecards/renderer/Util.java
```diff
@@ -349,6 +353,20 @@ public static<T extends BaseActionElement> T tryCastTo(BaseActionElement actionE
...
+     * Converts pixels to dps.
+     *
+     * @param context the Context for getting the resources
+     * @param px      dimension in pixels
+     * @return dimension in dps
+     */
+    public static @Dimension(unit = DP) int pixelToDp(@NonNull Context context, @Px int px) {
+        DisplayMetrics displayMetrics = context.getResources()
+            .getDisplayMetrics();
+        return Math.round(px / displayMetrics.density);
```
hggz, line 367: just a question to be resolved by you if you think we need to have a fallback value in case we get undefined values? (IE: 0, negative, etc)

ardlank: @hggz 
<= 0 will mean something went wrong getting the container's width. We will not use responsive layout in that cause.

Issue #23, opened 2023-10-19
[Adaptive Card iOS - Activity Update]: Voice Over does not convey the state for the 'Comment' button present under 'Activity Update' screen.


vagpt:
**Test Environment:**
Device: iPhone 11
iOS Version: 16.7.1
App Version: 1.0 (2.4.0-beta.23.10.15.1)
Screen Reader: Voice Over

**Repro Steps:**
1. Launch Adaptive Cards iOS application.
2. Home screen will appear.
3. Activate Voice over
4. Navigate to "V1.5" control and select it.
5. Navigate to 'Scenario' control and activate it.
6. Navigate to ' ActivateUpdate.Jsons' and double tap to activate it.
7. The Card will open.
8. Perform right swipe, navigate to 'Comment' button, and observe the issue.

**Actual Result:**
While navigating to the 'Comment' button first time, Voice over is not conveying the state information to the user. It is announcing only as **'Comment button'**.

**Observation:**
This issue is repro only first time when user navigates on the control. On navigating second time VO is announcing the state information.

**Expected Result:**
Voice Over should convey the state information to the user and when focus lands on it then vocie over should convey the information as **'Comment button collapsed double tap to expand it.'**.

**Note:**
1. Same issue is repro for the 'Set Due date' button.
2. This issue is repro throughout the application for the same type of control.

**User Experience:**
It will impact the screen reader user if the screen reader does not convey the state for the any control on the screen.

**WCAG Reference:**
https://www.w3.org/WAI/WCAG21/Understanding/info-and-relationships.html

[...]

PR #74, opened 2024-01-19
[iOS] Fix for accesibility label on expanded buttons

GabrielMedAlv:
Fix for accessibility issue #23

### DIFF
--- a/source/ios/AdaptiveCards/AdaptiveCards/AdaptiveCards/ACRShowCardTarget.mm
+++ b/source/ios/AdaptiveCards/AdaptiveCards/AdaptiveCards/ACRShowCardTarget.mm
@@ -84,6 +84,8 @@
     CGFloat showCardPadding = [_config getHostConfig]->GetSpacing().paddingSpacing;
 
     _adcView.backgroundColor = UIColor.clearColor;
+    
+    _button.accessibilityValue = NSLocalizedString(@"card collapsed", nil);
 
     if (@available(iOS 11.0, *)) {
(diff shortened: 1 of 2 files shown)


rahulpinjani1 (approved):

source/ios/AdaptiveCards/AdaptiveCards/AdaptiveCards/ACRShowCardTarget.mm
```diff
@@ -84,6 +84,8 @@ - (void)createShowCard:(NSMutableArray *)inputs superview:(UIView<ACRIContentHol
     CGFloat showCardPadding = [_config getHostConfig]->GetSpacing().paddingSpacing;
 
     _adcView.backgroundColor = UIColor.clearColor;
+    
+    _button.accessibilityValue = NSLocalizedString(@"card collapsed", nil);
```
JoshuaChingMSFT, line 88: nit: we could create a string variable to share between here and line 131: NSString *hint = hidden ? @"card expanded" : @"card collapsed";

JoshuaChingMSFT (approved):

source/ios/AdaptiveCards/AdaptiveCards/AdaptiveCards/ACRShowCardTarget.mm
GabrielMedAlv: Considering are different functions (initialization and toggle), and is a local variable just used once, could be better to be maintained as local, but I will check it.

PR #77, opened 2024-02-01
[iOS] fix voiceover duplicates for labels

JoshuaChingMSFT:
# Related Issue

Fixes Issue: https://github.com/microsoft/AdaptiveCards-Mobile/issues/24

# Description

A11Y voiceover announces the same piece of content twice when the accessibility label and accessibility value are the same. 

The fix is to clear the accessibility value if accessibility label and accessibility value are the same.

# How Verified

image

A11Y Inspector Before:
image

A11Y Inspector After:
image

### DIFF
--- a/source/ios/AdaptiveCards/AdaptiveCards/AdaptiveCards/ACRTextBlockRenderer.mm
+++ b/source/ios/AdaptiveCards/AdaptiveCards/AdaptiveCards/ACRTextBlockRenderer.mm
@@ -116,6 +116,11 @@
         lab.textContainer.lineBreakMode = NSLineBreakByTruncatingTail;
         lab.attributedText = content;
         lab.accessibilityLabel = content.string;
+        // if accessibility label is the same as accessibility value, clear accessibility value
+        // this prevents the same content from being repeated twice in voiceover
+        if (lab.accessibilityValue != nil && [lab.accessibilityValue isEqualToString:lab.accessibilityLabel]) {
+            lab.accessibilityValue = @"";
+        }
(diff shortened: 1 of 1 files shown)


GabrielMedAlv (approved):

GabrielMedAlv (approved):

source/ios/AdaptiveCards/AdaptiveCards/AdaptiveCards/ACRTextBlockRenderer.mm
```diff
@@ -116,6 +116,11 @@ - (UIView *)render:(UIView<ACRIContentHoldingView> *)viewGroup
         lab.textContainer.lineBreakMode = NSLineBreakByTruncatingTail;
         lab.attributedText = content;
         lab.accessibilityLabel = content.string;
+        // if accessibility label is the same as accessibility value, clear accessibility value
+        // this prevents the same content from being repeated twice in voiceover
+        if (lab.accessibilityValue != nil && [lab.accessibilityValue isEqualToString:lab.accessibilityLabel]) {
+            lab.accessibilityValue = @"";
```
hggz, line 122: yeah with a value and a label it can be repeated. This should work fine but maybe setting it to nil would imply that we don't want to use it more?

JoshuaChingMSFT: I initially tried that, but setting it to nil doesn't seem to clear the value (voice over still announces twice)

VikrantSingh01 (approved):

hggz (approved):

MengliElmendorf137 (approved):

rahulpinjani1 (approved):

PR #83, opened 2024-02-16
[iOS] - Fixed parsing detection for asterisk.

GabrielMedAlv:
Fixed parsing issue for text blocks that causes that \\* was not recognized as an asterisk on iOS. Added manual validation when the text is not an html.

Sample card payload:
`{
    "type": "AdaptiveCard",
    "version": "1.5",
    "body": [
        {
            "size": "medium",
            "text": "Change Contacts \\*SAMPLE TEXT\\* test",
            "weight": "regular",
            "type": "TextBlock"
        }
    ],
    "$schema": "http://adaptivecards.io/schemas/adaptive-card.json"
}
`

Before:

Simulator Screenshot - iPhone 15 - 2024-02-15 at 18 49 21

After:

Simulator Screenshot - iPhone 15 - 2024-02-15 at 18 45 38

### DIFF
--- a/source/ios/AdaptiveCards/AdaptiveCards/AdaptiveCards/UtiliOS.mm
+++ b/source/ios/AdaptiveCards/AdaptiveCards/AdaptiveCards/UtiliOS.mm
@@ -735,59 +735,46 @@ void buildIntermediateResultForText(ACRView *rootView, ACOHostConfig *hostConfig
         // MarkDownParser transforms text with MarkDown to a html string
         markdownString = markDownParser->TransformToHtml();
     }
-    NSString *parsedString = (markDownParser->HasHtmlTags()) ? [NSString stringWithCString:markdownString.c_str() encoding:NSUTF8StringEncoding] : [NSString stringWithCString:markDownParser->GetRawText().c_str() encoding:NSUTF8StringEncoding];
+    NSString *parsedString = [NSString stringWithCString:markdownString.c_str() encoding:NSUTF8StringEncoding];
 
-    if (markDownParser->HasHtmlTags() && ([parsedString containsString:@"\n"] || [parsedString containsString:@"\r"])) {
+    if ([parsedString containsString:@"\n"] || [parsedString containsString:@"\r"]) {
         parsedString = [parsedString stringByReplacingOccurrencesOfString:@"[\\n\\r]"
                                                                withString:@"<br>"
                                                                   options:NSRegularExpressionSearch
                                                                     range:NSMakeRange(0, [parsedString length])];
     }
-    // Added manual replace of \\* when the text is used directly, html parsing is exluding this case successfully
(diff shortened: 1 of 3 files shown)


source/ios/AdaptiveCards/AdaptiveCards/AdaptiveCards/UtiliOS.mm
```diff
@@ -743,7 +743,12 @@ void buildIntermediateResultForText(ACRView *rootView, ACOHostConfig *hostConfig
                                                                   options:NSRegularExpressionSearch
                                                                     range:NSMakeRange(0, [parsedString length])];
     }
-
+    // Added manual replace of \\* when the text is used directly, html parsing is exluding this case successfully
+    if (!markDownParser->HasHtmlTags() && [parsedString containsString:@"\\*"]) {
```
JoshuaChingMSFT, line 747: Is there a generic solution we can implement? This fix only accounts for * scenario.

For example, the same issue would occur for \\_SAMPLE TEXT\\_

GabrielMedAlv: That's a good question, in the case the text is html, the parsing process cover these scenarios. This is when is no html text and the original string is used. I could create a generic version to replace the "\\" with the next character after the lines. My only doubt if applies for all scenarios and are no cases when this could mean something else.

GabrielMedAlv: I have been doing some tests and, the rest of the scenarios are already covered, the "\\*" is the only scenario required because we don't have an alternative to put asterisks in a text block (single * is emphasis, ** is bolded and \* is not recognized by json parsing, even on web is not recognized). So, is ok to handle only this specific case. In the case you mentioned "\\_" is possible to display the "_" directly, not necessary any backslash

JoshuaChingMSFT: Have you tested this payload?

`{
    "type": "AdaptiveCard",
    "version": "1.4",
    "body": [
        {
            "size": "Medium",
            "text": "Change Contacts \\_SAMPLE TEXT\\_ test",
            "type": "TextBlock"
        }
    ],
    "$schema": "http://adaptivecards.io/schemas/adaptive-card.json"
}`

Desktop:
image

iOS:
image

GabrielMedAlv: Simplified solution to avoid visual diffs due to discussed change. After checking documentation, only 2 symbols ("*", "_") could take that behavior with double backslash.

ardlank (approved):

VikrantSingh01 (approved):

PR #118, opened 2024-04-09
[iOS]Add rounded corner and border support for some elements

wyAbhishek:
# Description

This PR intends to add support for showBorder and roundedCorners properties.  

## showBorder Property
If value of this property is true, then we will show 1pt border. Border colour will be same as container's style, taken from host config. We will be supporting showing borders on the following elements: 
1. Container 
2. ColumnSet 
3. Column 

Borders are already supported on the Table element. Default value is false.

## roundedCorners Property
If value of this property is true then we will show rounded corner for element. Corner radius is 5pt. 
We will be supporting rounded corners on the following elements: 

1. Container 
2. ColumnSet 
3. Column 
4. Table 

We are adding rounded corner support for Image element also. But it will be enabled through existing ImageStyle property. So, style of Image can be Default, Person and now RoundedCorners. 

# Sample Card
Added some mock elements with showBorder and roundedCorner properties set to true. 

Adding some screenshots for these changes below:
Simulator Screenshot - iPhone 15 Pro - 2024-04-09 at 09 40 03
Simulator Screenshot - iPhone 15 Pro - 2024-04-10 at 10 10 00

Simulator Screenshot - iPhone 15 Pro - 2024-04-09 at 09 47 44
Simulator Screenshot - iPhone 15 Pro - 2024-04-09 at 09 41 42


source/ios/AdaptiveCards/AdaptiveCards/AdaptiveCards/ACRColumnRenderer.mm
```diff
@@ -113,4 +115,26 @@ - (void)configUpdateForUIImageView:(ACRView *)rootView acoElem:(ACOBaseCardEleme
...
 }
 
+- (void)configureBorderForElement:(ACOBaseCardElement *)acoElem container:(ACRContentStackView *)container config:(ACOHostConfig *)acoConfig
+{
+    std::shared_ptr<BaseCardElement> elem = [acoElem element];
+    std::shared_ptr<Column> containerElem = std::dynamic_pointer_cast<Column>(elem);
+    bool shouldShowBorder = containerElem->GetShowBorder();
+    if(shouldShowBorder)
+    {
+        container.layer.borderWidth = 1;
```
wyAbhishek, line 125: Should borderWidth(1pt) and cornerRadius(5pt) come from host config? Or any central constant file needed?
@ardlank @GabrielMedAlv

GabrielMedAlv: I think could be a good idea to have them on host config, do you agree @ardlank, to have it on android on host config too?

GabrielMedAlv (approved):

source/ios/AdaptiveCards/AdaptiveCards/AdaptiveCards/ACRColumnRenderer.mm
ardlank: Yes, there should be a default value and the property can be overridden by host config. @wyAbhishek @GabrielMedAlv

wyAbhishek: Looks like there is an issue with this.

1. We already have borderThickness with default value 0 [here](https://github.com/microsoft/AdaptiveCards-Mobile/blob/a359afded0f946a5c386b81c94a065d59157869f/source/shared/cpp/ObjectModel/HostConfig.h#L199C1-L200C1).  If we change default value then all renderers using ACRContentView will start showing border. 
2. Image can't have this because image doesn't have container style and rounded corner is based on ImageStyle.
3. From the Code comment [here](https://github.com/microsoft/AdaptiveCards-Mobile/blob/a359afded0f946a5c386b81c94a065d59157869f/source/shared/cpp/ObjectModel/HostConfig.h#L199C1-L200C1), it seems that borderThickness is not consumed in Android. 

IMO, we should decide clear schema for HostConfig and then move constants accordingly. I can do that in follow up PRs if we decide to change anything.

dclaux:
Looks good to me. I don't seem to be able to approve though.

source/shared/cpp/ObjectModel/Enums.cpp
```diff
@@ -269,7 +271,8 @@ namespace AdaptiveCards
     DEFINE_ADAPTIVECARD_ENUM(ImageStyle, {
             {
                 {ImageStyle::Default, "default"},
-                {ImageStyle::Person, "person"}
+                {ImageStyle::Person, "person"},
+                {ImageStyle::RoundedCorners, "roundedCorners"}
```
ardlank, line 275: Why is roundedCorners a boolean everywhere but a style for images?

wyAbhishek: It is coming from spec:
"We will also support rounded corners on Image elements, using the existing style property:  Default, Person and now RoundedCorners "

ardlank (approved):
Review the shared changes and looks good to me. @GabrielMedAlv please review the iOS changes

PR #120, opened 2024-04-16
[Android] Add border and roundedCorner support for different elements

singhmanp0707:
# Description

This PR contains the following changes:
1. Add border for the elements - Column, ColumnSet and Container, if the showBorder property is passed as true
2. Add roundedCorners for the elements - Column, ColumnSet, Container and Table, if the roundedCorners property is passed as true.
3. Added rounded corner support for image element if the image style property is set as "roundedCorners".

# Sample Card

Screenshot 2024-04-16 at 5 14 04 PM

Screenshot 2024-04-16 at 5 15 45 PM

Screenshot 2024-04-16 at 5 16 47 PM

Screenshot 2024-04-17 at 1 44 19 PM

Screenshot 2024-04-17 at 1 45 45 PM


source/android/adaptivecards/src/main/java/io/adaptivecards/renderer/readonly/TableRenderer.kt
```diff
@@ -54,6 +56,24 @@ object TableRenderer : BaseCardElementRenderer() {
...
+                        this.setStroke(Util.dpToPixels(tableLayout.context, 1f), Color.parseColor(hostConfig.GetBorderColor(computedGridStyle)))
+                    }
+                } else {
+                    tableLayout.background = GradientDrawable().apply {
+                        this.cornerRadius = cornerRadius
+                        this.setStroke(Util.dpToPixels(tableLayout.context, 1f), Color.parseColor(hostConfig.GetBorderColor(computedGridStyle)))
+                    }
+                }
+            }
+            tableLayout.clipToOutline = true
```
singhmanp0707, line 74: Will move it inside the ShowGridLines condition check.

ardlank:
Looks like your swig version is 4.2.1. You need to downgrade it to 4.0.2.
 
In the past, when we made the SDK with 4.2.1, the Teams apk size increased too much.

PR #128, opened 2024-04-29
Added ConditionallyEnabled property in SubmitAction and ExecuteAction 

prpatwa:
# Description

For any Action.Submit or Action.Execute element in adaptive card if ConditionallyEnabled value is set true, then the button is enabled/disabled based on the other input fields. 

If this property is true, InputWatcher is added as input change observer to all the input Fields via InputHandlers. If any of the input is valid button state changes to Enable and if all the input fields are invalid button state is kept disabled.

# Sample Card

Added a sample json - ConditionallyEnabledAction.json

# How Verified

How you verified the fix, including one or all of the following:

https://github.com/microsoft/AdaptiveCards-Mobile/assets/68814139/c0123826-39c8-4cdc-a188-05a8caccaec9


ardlank (approved):

karthikgopal36 (approved):

hggz (approved):
LGTM

PR #204, opened 2024-08-19
[Android] [Bug] Fix  - Unordered list followed by ordered list not working correclty

karthikgopal36:
# Description

If a text contains unordered list followed by ordered list, the bullet point is not showing and replaced with numbers.

# Sample Card

```
                {
			"type": "TextBlock",
			"text": "1. Numbered\r- List\r",
			"wrap": true
		}
```

# How Verified
verified with the card

Before             |  After
:-------------------------:|:-------------------------:
screenshot-1724048798505 |  screenshot-1724048757508

### DIFF
--- a/source/android/adaptivecards/src/main/java/io/adaptivecards/renderer/readonly/RendererUtil.java
+++ b/source/android/adaptivecards/src/main/java/io/adaptivecards/renderer/readonly/RendererUtil.java
@@ -208,6 +208,12 @@ public class RendererUtil
         {
             boolean tagWasHandled = false;
 
+            if ((tag.equals("ol") && !opening))
+            {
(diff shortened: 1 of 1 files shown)


GabrielMedAlv:
Thanks for fixing this @karthikgopal36 . do you know if this is android specific or might be required to be handled on iOS too?

GabrielMedAlv (approved):

ardlank (approved):

karthikgopal36:
> Thanks for fixing this @karthikgopal36 . do you know if this is android specific or might be required to be handled on iOS too?

@GabrielMedAlv  this is android only issue. iOS works fine.

PR #220, opened 2024-08-23
fix: sanitize unescaped problematic url regex strings

hggz:
# Related Issue

N/A

# Description

This fixes faulty json processing of unescaped regex following the below url pattern for adaptive cards on iOS 

# Sample Broken JSON

"regex": "^([a-zA-Z0-9._%+-]+@outokumpu\.com)(,[a-zA-Z0-9._%+-]+@outokumpu\.com)*$",

# Sample Expected JSON

"regex": "^([a-zA-Z0-9._%+-]+@outokumpu\\.com)(,[a-zA-Z0-9._%+-]+@outokumpu\\.com)*$",

# How Verified

I validated the json payload in the `samples/v1.4/Tests/ActionExecute.With.Validation.json` substituting the broken json and attempting to render the card in the ACVisuzlier.

# Considerations

This technically fixes it but there are multiple approaches. We can anticipate an error and then retroactively attempt to apply this pattern as a fallback. We can also double down and add extra JSON parsing correctness using the Foundation JSON Serialization library and report on errors that way to attempt to create a more graceful solution.

### DIFF
--- a/source/ios/AdaptiveCards/AdaptiveCards/AdaptiveCards/ACOAdaptiveCard.mm
+++ b/source/ios/AdaptiveCards/AdaptiveCards/AdaptiveCards/ACOAdaptiveCard.mm
@@ -56,41 +56,115 @@ using namespace AdaptiveCards;
     return _inputs;
 }
 
-+ (ACOAdaptiveCardParseResult *)fromJson:(NSString *)payload;
-{
++ (NSString *)correctInvalidJsonEscapes:(NSString *)payload {
+    NSError *regexError = nil;
+    NSRegularExpression *regex = [NSRegularExpression regularExpressionWithPattern:@"(\\\\+\\.)"
+                                                                           options:0
+                                                                             error:&regexError];
+    if (regexError != nil) {
+        NSLog(@"Regex Error: %@", regexError.localizedDescription);
+        return payload; // Return the original payload if regex creation fails
+    }
+    
+    NSMutableString *mutablePayload = [payload mutableCopy];
+
+    // Replace matches with the correct number of escaped backslashes
+    [regex enumerateMatchesInString:payload
+                            options:0
+                              range:NSMakeRange(0, payload.length)
+                         usingBlock:^(NSTextCheckingResult *match, NSMatchingFlags flags, BOOL *stop) {
+        if (match.range.length > 0) {
+            NSString *matchedString = [payload substringWithRange:match.range];
+            NSString *replacementString = [matchedString stringByReplacingOccurrencesOfString:@"\\" withString:@"\\\\"];
+            [mutablePayload replaceCharactersInRange:match.range withString:replacementString];
+        }
+    }];
+
+    return [mutablePayload copy];
+}
+
++ (BOOL)isValidJson:(NSString *)jsonString error:(NSError **)error {
+    NSData *jsonData = [jsonString dataUsingEncoding:NSUTF8StringEncoding];
+    if (!jsonData) {
+        return NO;
+    }
+
+    [NSJSONSerialization JSONObjectWithData:jsonData options:0 error:error];
+    return (*error == nil);
+}
+
+
++ (ACOAdaptiveCardParseResult *)fromJson:(NSString *)payload {
     const std::string g_version = "1.6";
     ACOAdaptiveCardParseResult *result = nil;
     if (payload) {
-        try {
-            ACOAdaptiveCard *card = [[ACOAdaptiveCard alloc] init];
-            NSString *processedPayload = [payload stringByReplacingOccurrencesOfString:@"\\." withString:@"\\\\."];
-            std::shared_ptr<ParseResult> parseResult = AdaptiveCard::DeserializeFromString(std::string([processedPayload UTF8String]), g_version);
-            NSMutableArray *acrParseWarnings = [[NSMutableArray alloc] init];
-            std::vector<std::shared_ptr<AdaptiveCardParseWarning>> parseWarnings = parseResult->GetWarnings();
-            for (const auto &warning : parseWarnings) {
-                ACRParseWarning *acrParseWarning = [[ACRParseWarning alloc] initWithParseWarning:warning];
-                [acrParseWarnings addObject:acrParseWarning];
-            }
-            card->_adaptiveCard = parseResult->GetAdaptiveCard();
-            if (card && card->_adaptiveCard) {
-                card->_refresh = [[ACORefresh alloc] init:card->_adaptiveCard->GetRefresh()];
-                card->_authentication = [[ACOAuthentication alloc] init:card->_adaptiveCard->GetAuthentication()];
-            }
-            result = [[ACOAdaptiveCardParseResult alloc] init:card errors:nil warnings:acrParseWarnings];
-        } catch (const AdaptiveCardParseException &e) {
-            // converts AdaptiveCardParseException to NSError
-            ErrorStatusCode errorStatusCode = e.GetStatusCode();
-            NSInteger errorCode = (long)errorStatusCode;
-            NSBundle *adaptiveCardsBundle = [[ACOBundle getInstance] getBundle];
-            NSString *localizedFormat = NSLocalizedStringFromTableInBundle(@"AdaptiveCards.Parsing", nil, adaptiveCardsBundle, "Parsing Error Messages");
-            NSString *objectModelErrorCodeInString = [NSString stringWithCString:ErrorStatusCodeToString(errorStatusCode).c_str() encoding:NSUTF8StringEncoding];
-            NSDictionary<NSErrorUserInfoKey, id> *userInfo = @{NSLocalizedDescriptionKey : [NSString localizedStringWithFormat:localizedFormat, objectModelErrorCodeInString]};
-            NSError *parseError = [NSError errorWithDomain:ACRParseErrorDomain
-                                                      code:errorCode
-                                                  userInfo:userInfo];
-            NSArray<NSError *> *errors = @[ parseError ];
+        NSError *jsonError = nil;
 
-            result = [[ACOAdaptiveCardParseResult alloc] init:nil errors:errors warnings:nil];
+        // First, check if the JSON is valid
+        if (![self isValidJson:payload error:&jsonError]) {
+            NSLog(@"Initial JSON Deserialization Error: %@", jsonError.localizedDescription);
+            jsonError = nil;
+
+            // Attempt to fix the JSON using regex replacement
+            NSString *processedPayload = [self correctInvalidJsonEscapes:payload];
+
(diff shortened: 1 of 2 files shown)


ardlank (approved):

source/ios/AdaptiveCards/AdaptiveCards/AdaptiveCards/ACOAdaptiveCard.mm
```diff
@@ -63,7 +63,8 @@ + (ACOAdaptiveCardParseResult *)fromJson:(NSString *)payload;
     if (payload) {
         try {
             ACOAdaptiveCard *card = [[ACOAdaptiveCard alloc] init];
-            std::shared_ptr<ParseResult> parseResult = AdaptiveCard::DeserializeFromString(std::string([payload UTF8String]), g_version);
+            NSString *processedPayload = [payload stringByReplacingOccurrencesOfString:@"\\." withString:@"\\\\."];
```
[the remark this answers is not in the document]
hggz: if the json payload has double slashes (or more) this isn't guaranteed to work. I was thinking perhaps we can do some json parsing recovery code. I wasn't sure if we should include it now. 

If the existing logic works, continue as usual. Otherwise, try to sanitize this payload similarly. (Like below)

```
+ (ACOAdaptiveCardParseResult *)fromJson:(NSString *)payload {
    const std::string g_version = "1.6";
    ACOAdaptiveCardParseResult *result = nil;
    if (payload) {
        NSData *jsonData = [payload dataUsingEncoding:NSUTF8StringEncoding];
        NSError *jsonError = nil;

        // Initial attempt to deserialize JSON data
        id jsonObject = [NSJSONSerialization JSONObjectWithData:jsonData options:0 error:&jsonError];
        
        // If there's an error, try to fix the payload
        if (jsonError != nil) {
            NSLog(@"Initial JSON Deserialization Error: %@", jsonError.localizedDescription);
            jsonError = nil;
            // Apply the string replace fix for common escape issues
            NSString *processedPayload = [payload stringByReplacingOccurrencesOfString:@"\\." withString:@"\\\\."];
            jsonData = [processedPayload dataUsingEncoding:NSUTF8StringEncoding];

            // Attempt to deserialize the corrected JSON data
            jsonObject = [NSJSONSerialization JSONObjectWithData:jsonData options:0 error:&jsonError];
        }

        if (jsonError != nil) {
            return result; // Return nil result if JSON is invalid
        }

        // Continue with logic if JSON is valid
        if (jsonObject != nil) {
            // Re-serialize the object to JSON data
            jsonData = [NSJSONSerialization dataWithJSONObject:jsonObject options:0 error:&jsonError];
            if (jsonError == nil && jsonData != nil) {
                // Convert back to NSString
                NSString *escapedPayload = [[NSString alloc] initWithData:jsonData encoding:NSUTF8StringEncoding];
                
                try {
                    ACOAdaptiveCard *card = [[ACOAdaptiveCard alloc] init];
                    std::shared_ptr<ParseResult> parseResult = AdaptiveCard::DeserializeFromString(std::string([escapedPayload UTF8String]), g_version);
                    NSMutableArray *acrParseWarnings = [[NSMutableArray alloc] init];
                    std::vector<std::shared_ptr<AdaptiveCardParseWarning>> parseWarnings = parseResult->GetWarnings();
                    for (const auto &warning : parseWarnings) {
                        ACRParseWarning *acrParseWarning = [[ACRParseWarning alloc] initWithParseWarning:warning];
                        [acrParseWarnings addObject:acrParseWarning];
                    }
                    card->_adaptiveCard = parseResult->GetAdaptiveCard();
                    if (card && card->_adaptiveCard) {
                        card->_refresh = [[ACORefresh alloc] init:card->_adaptiveCard->GetRefresh()];
                        card->_authentication = [[ACOAuthentication alloc] init:card->_adaptiveCard->GetAuthentication()];
                    }
                    result = [[ACOAdaptiveCardParseResult alloc] init:card errors:nil warnings:acrParseWarnings];
                } catch (const AdaptiveCardParseException &e) {
                    // Converts AdaptiveCardParseException to NSError
                    ErrorStatusCode errorStatusCode = e.GetStatusCode();
                    NSInteger errorCode = (long)errorStatusCode;
                    NSBundle *adaptiveCardsBundle = [[ACOBundle getInstance] getBundle];
                    NSString *localizedFormat = NSLocalizedStringFromTableInBundle(@"AdaptiveCards.Parsing", nil, adaptiveCardsBundle, "Parsing Error Messages");
                    NSString *objectModelErrorCodeInString = [NSString stringWithCString:ErrorStatusCodeToString(errorStatusCode).c_str() encoding:NSUTF8StringEncoding];
                    NSDictionary<NSErrorUserInfoKey, id> *userInfo = @{NSLocalizedDescriptionKey : [NSString localizedStringWithFormat:localizedFormat, objectModelErrorCodeInString]};
                    NSError *parseError = [NSError errorWithDomain:ACRParseErrorDomain
                                                              code:errorCode
                                                          userInfo:userInfo];
                    NSArray<NSError *> *errors = @[ parseError ];

                    result = [[ACOAdaptiveCardParseResult alloc] init:nil errors:errors warnings:nil];
                }
            } else {
                // Handle JSON serialization error
                NSLog(@"JSON Serialization Error: %@", jsonError.localizedDescription);
            }
        }
    }
    return result;
}
```

but if we start supporting json recovery this way, we may want to formalize it and add these recovery rules elsewhere.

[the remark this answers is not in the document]
rahulpinjani1: can we include a regex check here instead of to see if we can parse both double, 4 backslashes or even more using NSRegularExpression?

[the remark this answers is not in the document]
hggz: Good call. I ended up splitting the proposed json validation logic above into a helper function and tested it against variable amounts of backslashes. This works. I tested it against the regex pattern in

`/samples/v1.4/Tests/Action.Execute.With.Validation` 

Should I add another sample test?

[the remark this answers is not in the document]
rahulpinjani1: Thanks for addressing this, if the existing test covers the flow then we are good else we can add one just to be safe. :)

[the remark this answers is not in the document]
rahulpinjani1: Thanks for addressing this, if the existing test covers the flow then we are good else we can add one just to be safe. :)

rahulpinjani1 (approved):

source/ios/AdaptiveCards/AdaptiveCards/AdaptiveCards/ACOAdaptiveCard.mm
```diff
@@ -56,40 +56,115 @@ - (NSData *)inputs
...
 
-+ (ACOAdaptiveCardParseResult *)fromJson:(NSString *)payload;
-{
++ (NSString *)correctInvalidJsonEscapes:(NSString *)payload {
+    NSError *regexError = nil;
+    NSRegularExpression *regex = [NSRegularExpression regularExpressionWithPattern:@"(\\\\+\\.)"
+                                                                           options:0
+                                                                             error:&regexError];
+    if (regexError != nil) {
+        NSLog(@"Regex Error: %@", regexError.localizedDescription);
```
JoshuaChingMSFT, line 65: Do we need to be logging this? We should consider making this more "informational" sounding. It seems like hitting this line of code isn't necessarily an error.

source/ios/AdaptiveCards/AdaptiveCards/AdaptiveCards/ACOAdaptiveCard.mm
```diff
@@ -56,40 +56,115 @@ - (NSData *)inputs
...
+        // First, check if the JSON is valid
+        if (![self isValidJson:payload error:&jsonError]) {
+            NSLog(@"Initial JSON Deserialization Error: %@", jsonError.localizedDescription);
+            jsonError = nil;
+
+            // Attempt to fix the JSON using regex replacement
+            NSString *processedPayload = [self correctInvalidJsonEscapes:payload];
+
+            // Check if the corrected JSON is valid
+            if (![self isValidJson:processedPayload error:&jsonError]) {
```
JoshuaChingMSFT, line 112: Why do we need to check again after post processing? Does this mean that the correctInvalidJsonEscapes method could potentially create invalid jsons?

GabrielMedAlv:
My concern is if the fix could increase parsing time for cards in general.

source/ios/AdaptiveCards/AdaptiveCards/AdaptiveCards/ACOAdaptiveCard.mm
hggz: Technically this could be an error if there is malformed regex, and it would be caught with the next check under 'isValidJson' 

I can remove this log and do something more informal sounding like you suggest though

source/ios/AdaptiveCards/AdaptiveCards/AdaptiveCards/ACOAdaptiveCard.mm
hggz: In regards to @GabrielMedAlv  concern about increasing parsing times for AC in general, I circumvent this by only doing this check if we detect an error in the original json payload. The intended flow is:

- Check if the AC json is valid
- if the AC json isn't valid, attempt to fix it by fixing faulty Regex cases
- Validate fixed json
- continue as normal

before this the flow was:
- If the AC Json isn't valid, do nothing and card is broken

hggz:
> My concern is if the fix could increase parsing time for cards in general.

I commented on this concern above in josh's comment

source/ios/AdaptiveCards/AdaptiveCards/AdaptiveCards/ACOAdaptiveCard.mm
JoshuaChingMSFT: Sorry, I misread the code, you are right. Let's keep the log 👍

source/ios/AdaptiveCards/AdaptiveCards/AdaptiveCards/ACOAdaptiveCard.mm
hggz: RE: In regards to performance the new intended flow is:

- Check if the AC json is valid
- if the AC json isn't valid, attempt to fix it by fixing faulty Regex cases
- continue as normal

(we don't need to validate the fixed json, it either works now or fails how we had it before)

GabrielMedAlv (approved):

PR #222, opened 2024-08-26
Flow layout support for iOS Phase 1

wyAbhishek:
# Support for Flow Layout Container

- This PR adds support for the Flow Layout container for elements of type AdaptiveCard, Container, Column and Table Cell. In Phase 1, we are distributing width of of parent equally in column set and table. 
- This PR introduces a feature flag resolver interface that allows the host to implement and register it, enabling control of features through ECS settings.

A new **_layouts_** property is introduced on container elements. The value is an array of Layout objects. The layout objects look like this:

Screenshot 2024-08-23 at 5 03 23 PM

The layout to apply is determined the following way: 

- Look for the first layout in the list whose targetWidth property matches the current host width. If one is found, apply it 

- Otherwise, look for a layout that doesn’t have a targetWidth defined. If one is found, apply it 

- Otherwise, apply a default Layout.Stack layout and ignore other layouts in the list 

# Spec for Layout.Flow

Screenshot 2024-08-23 at 4 57 38 PM

Sample Card:
Simulator Screenshot - iPhone 15 Pro - 2024-08-26 at 16 45 54


hggz:
Reminder comment to fix SPM support with new header additions.

PR #231, opened 2024-09-03
Fix for no area specified in Area Grid

wyAbhishek:
If there is no area specified in area grid, then it should behave as Flow layout. This case is not handled currently. 

To fix this we added check for no area in case of Area layout. We convert to flow layout in this case.  These changes are done in all layouts: column, container, TableCell and card


prpatwa (approved):
Nit - Can you please add screenshot here.

PR #232, opened 2024-09-04
[Android + iOS]: Allow configuration of AC Fluent icon CDN paths for different environments

singhmanp0707:
# Description

### Android changes
Currently, Fluent icon assets for AC are served from this unique CDN path:
https://res-1.cdn.office.net/assets/fluentui-react-icons/2.0.226/,
But The base path for this CDN should actually be different based on the environment. This PR adds another method in IFeatureFlagResolver to fetch the ECS settings as String for a key. This method will be used to fetch the FluentIconCdnRoot and FluentIconCdnPath from ECS and use those to create a Cdn URL to fetch and render fluent icons.

The host will register its featureFlagResolver and will implement the getEcsSettingAsString method to fetch the values from the ECS.

### iOS changes
Currently, Fluent icon assets for AC are served from this unique CDN path:
https://res-1.cdn.office.net/assets/fluentui-react-icons/2.0.226/,
But The base path for this CDN should actually be different based on the environment. This PR uses FeatureFlagResolver to fetch root CDN URL from host application. 

if host doesn't provide value for flag "fluentIconCdnURL", then we will use existing default url. 

# Sample Card

FluentIcon

TeamsFluentIcon

compoundButton

TeamsCompoundButton


wyAbhishek (approved):

ihiteshkumar (approved):
LGTM

prpatwa (approved):

PR #233, opened 2024-09-05
[iOS] Flow layout phase2 

wyAbhishek:
# Flow Layout Phase 2

In flow layout phase-1 we added a width distribution logic which always distributes width equally amongst all it's childern. Basically, if available width is x and number of items in any container is y then each item will have x/y width. 

In phase2 , we are more complex width distribution logic. Following is the new logic:

### Container and Card:
In Container and Card, we will check if flow layout applies. If applies than each child will get item width specified in layout.
Otherwise all items will take parents width

### Column and TableCell:
Here we can have different type of widths provided for container. For example: Column can have "auto", "stretch",  fixed pixel width or relative width. 
Based on value of width, the available width will be divided.


source/ios/AdaptiveCards/AdaptiveCards/AdaptiveCards/ACRLayoutHelper.mm
```diff
@@ -264,5 +286,239 @@ - (BOOL)shouldUseNewLayoutForView:(ACRView *)rootView
...
+ * This methods distributes available width amongst all items of given container
+ */
+- (float)containerWidthDistributor:(std::shared_ptr<Container> const)container
+                    availableWidth:(float)availableWidth
+                     andHostConfig:(ACOHostConfig *)config
+{
+    std::shared_ptr<Layout> final_layout = [self layoutToApplyFrom:container->GetLayouts() andHostConfig:config];
+    if(final_layout->GetLayoutContainerType() == LayoutContainerType::Flow)
+    {
+        std::shared_ptr<FlowLayout> flow_layout = std::dynamic_pointer_cast<FlowLayout>(final_layout);
```
guptaabhishe, line 299: we need to dynamically cast here, can we not use any better approach like deserialize

guptaabhishe (approved):
Looks good to me, left a comment please check

ihiteshkumar (approved):

PR #234, opened 2024-09-05
[Android + iOS]: Fix: Rating element may not render if max value is missing

wyAbhishek:
# Related Issue
For rating element and rating input element, there was logic to validate that max is always greater than value. 
This logic is not present in web. Which leads to card rendering on web client but not rendering in mobile.

# Fix
We removed all parsing failure logic from rating elements.  Max will always be more or equal to value.


source/shared/cpp/ObjectModel/RatingInput.cpp
```diff
@@ -101,9 +101,14 @@ std::shared_ptr<BaseCardElement> RatingInputParser::Deserialize(ParseContext& co
...
-    if (ratingInput->GetMax() == 0 || (ratingInput->GetValue() > ratingInput->GetMax()))
+    if (ratingInput->GetMax() == 0 && ratingInput->GetValue() != 0)
     {
-        throw AdaptiveCardParseException(ErrorStatusCode::InvalidPropertyValue, "Max value of Rating Input should not be smaller than rating value");
+        ratingInput->SetMax(ratingInput->GetValue());
+    }
+    
+    if(ratingInput->GetMax() == 0 && ratingInput->GetValue() == 0)
+    {
+        throw AdaptiveCardParseException(ErrorStatusCode::InvalidPropertyValue, "In rating label Both max and value can't be zero");
```
singhmanp0707, line 111: nit: The exception description should be "In rating input, both max and value can't be zero"

singhmanp0707 (approved):

ihiteshkumar (approved):

PR #239, opened 2024-09-09
[Android] Adding support for AreaGrid Layout Container

prpatwa:
Adding support for AreaGrid Layout Container

# Description
This PR implements AreaGrid Layout container functionality for AdaptiveCard, Container, Column, and Table Cell elements.
A new layouts property has been added to container elements. This property accepts an array of Layout objects, which are defined as follows:

image

Properties for AreaGrid layout:

image

# Sample Card
ContainerAreaGrid, AdaptiveCardAreaGrid, ColumnAreaGrid, TableAreaGrid
image

# Card
image

All the test cases added in this [document](https://microsoftapc-my.sharepoint.com/:w:/g/personal/prpatwa_microsoft_com/EdAFMBJVuHRHl7PRLTznhEUBhKC2QMTgev9Je2nzB_xoCQ?e=e6tVcE)

Recording of all the test cases
https://github.com/user-attachments/assets/83870c79-f74f-401e-b10d-b80b31c66e88


ardlank (approved):

source/shared/cpp/ObjectModel/BaseCardElement.cpp
```diff
@@ -228,4 +238,6 @@ void BaseCardElement::DeserializeBaseProperties(ParseContext& context, const Jso
     element->SetSeparator(ParseUtil::GetBool(json, AdaptiveCardSchemaKey::Separator, false));
     element->SetSpacing(ParseUtil::GetEnumValue<Spacing>(json, AdaptiveCardSchemaKey::Spacing, Spacing::Default, SpacingFromString));
     element->SetAreaGridName(ParseUtil::GetString(json, AdaptiveCardSchemaKey::AreaGridName, "", false));
```
wyAbhishek, line 240: We are already setting default value to empty string here. So we will always get a value for area grid.  Do we need another method?

prpatwa: optional string is not supported in java. swig convert it to this type SWIGTYPE_p_std__optionalT_std__string_t that we can't use.

prpatwa: Changed it to non optional String with default value.

prpatwa: Reverted "non optional String with default value."  as it was breaking iOS build. Will take it as a follow up item.

wyAbhishek (approved):

PR #291, opened 2025-02-10
[iOS] Remove Button xibs.

DongyuZhao:
# Related Issue

Per Office build policy, xib files should be removed from the project.

# Description

This PR contains the OBJC implementation for ACRButton and ACRButtonExpandable.

It also removed unused chevronup and chevrondown svg files.

# How Verified

This change is an inline replacement and all the existing test and verification for these files should remain the same.


nemowon (approved):
if possible, could you attach some screenshot during verify the UI elements

PR #292, opened 2025-02-10
[iOS] Remove the unnecessary ACRDatePicker.xib

DongyuZhao:
# Related Issue

Per Office build policy, xib files should be removed from the project.

# Description

This PR remove the unnecessary ACRDatePicker.xib as it can be covered by the UIDatePicker logic.

# How Verified

This change is an inline replacement and all the existing test and verification for these files should remain the same.


nemowon (approved):
looks good

PR #297, opened 2025-02-11
[iOS]Add missing register case for ACRView

nemowon:
# Related Issue

Add missing register case for ACRView

# Description

Add missing register case for ACRView


nemowon:
@GabrielMedAlv  
the logic should be same as prod env,
but please help confirm that the explicit case is desired in case I accidentally remove some meaningful case warning, Thanks!

DongyuZhao (approved):

GabrielMedAlv (approved):

PR #304, opened 2025-02-16
[iOS] Bump minimum target to iOS 15.0

DongyuZhao:
# Related Issue

Bump to iOS 15.0 to reduce the unnecessary size bloating brought by external resources.

ContentEdgeInsets related API has been deprecated since iOS 15.0, to calm down compiler, we migrate ACRButton to UIButtonConfiguration in this pull request. Since the new API do not support customized UIImageView, we did some refactor to ACRSVGImageView to make the SVG download and render methods can be access without using an instance of ACRSVGImageView.

# How Verified

This is an inline replacement of the external resources with the SF symbols alongside with an API migration, all the verification should remain the same.

Run a thorough check for each scenarios in visualizer, no visual change observed.


nemowon (approved):

GabrielMedAlv (approved):

gauravkeshre:
Ideally, we should revert the file to previous commit SHA instead of redoing the changes. But given the urgency this is ok. 
Looks good to me. Let us merge it.

PR #376, opened 2025-05-14
Added a host config value to show split button

P-Harika:
If hostconfig contains `isSplitButtonEnabled` boolean as set under `actions`, then we show chevron down icon and parse menuActions.


source/shared/cpp/ObjectModel/BaseActionElement.cpp
```diff
@@ -257,7 +259,7 @@ void BaseActionElement::DeserializeBaseProperties(ParseContext& context, const J
     auto themedUrls = ParseUtil::GetElementCollectionOfSingleType<ThemedUrl>(context, json, AdaptiveCardSchemaKey::ThemedIconUrls, ThemedUrl::Deserialize, false);
     element->m_themedIconUrls = std::move(themedUrls);
 
-    if (IsSplitActionSupported(element->GetElementType())) {
+    if (IsSplitActionSupported(element->GetElementType()) && BaseActionElement::isSplitActionEnabledForHost) {
```
vntGitMicrosoft, line 262: Android Teams is not passing isSplitButtonEnabled in host config. As it is in shared layer, Won't it stop working for Teams android in that case.

source/ios/AdaptiveCards/AdaptiveCards/AdaptiveCards/ACRRenderer.mm
```diff
@@ -60,13 +61,15 @@ + (ACRRenderResult *)render:(ACOAdaptiveCard *)card
...
 {
+    BaseActionElement::SetIsSplitActionEnabledForHost([config isSplitButtonEnabled]);
     return [ACRRenderer render:card config:config widthConstraint:width delegate:nil theme:theme];
 }
 
 // This interface is exposed to outside, and returns ACRRenderResult object
 // This object contains a viewController instance which defer rendering adaptiveCard until viewDidLoad is called.
 + (ACRRenderResult *)renderAsViewController:(ACOAdaptiveCard *)card config:(ACOHostConfig *)config frame:(CGRect)frame delegate:(id<ACRActionDelegate>)acrActionDelegate theme:(ACRTheme)theme
 {
+    BaseActionElement::SetIsSplitActionEnabledForHost([config isSplitButtonEnabled]);
```
meethjain, line 72: is there a way to centralise this?

source/shared/cpp/ObjectModel/BaseActionElement.cpp
P-Harika: Discussed offline.
Teams Android need to make changes, which can be part of next AC SDK version integration.

source/ios/AdaptiveCards/ADCIOSVisualizer/resources/filebrowserHostConfig.json
```diff
@@ -215,7 +215,8 @@
 		},
 		"preExpandSingleShowCardAction": false,
 		"actionsOrientation": "Horizontal",
-		"actionAlignment": "Left"
+		"actionAlignment": "Left",
+		"isSplitButtonEnabled": true
```
wyAbhishek, line 219: I think enable/disable is not part of Host Config. Default UI related values are read here.

P-Harika: I see other usages like `isSubtle`, `wrap` where UI changes are handled from config values. The reason for this json value is that split button needs host implementation of on click of split button - Like Teams show bottom sheet. Without this check, Other hosts might see the chevron down icon but since they wouldn't have implemented this, the button looks like split button. To avoid this, value in host config is needed.

meethjain: Is there another way to pass this in iOS, we were doing this by sending it as part of Registration components in Android, is there something similar that can be used?

@wyAbhishek

wyAbhishek: There should be similar implementation in iOS as well. Check for flag name "isFlowLayoutContainerEnabled" for example.

source/ios/AdaptiveCards/ADCIOSVisualizer/ADCIOSVisualizer/ACRCustomFeatureFlagResolver.m
```diff
@@ -27,6 +29,10 @@ - (BOOL)boolForFlag:(NSString *)flag
     {
         return YES;
     }
+    if([flag isEqual:isSplitButtonEnabledKey])
```
gauravkeshre, line 32: We check for case insensitive string comparison. Also use `isEqualToString` instead.

gauravkeshre (approved):

source/ios/AdaptiveCards/ADCIOSVisualizer/resources/filebrowserHostConfig.json
P-Harika: Changes updated as per offline discussion, using ACRCustomFeatureFlagResolver

meethjain (approved):

PR #390, opened 2025-06-11
[Android] Image fitMode implemented

vntGitMicrosoft:
#### Description
When an image is sized using both its width and height properties, it is stretched to cover its bounding box. This can cause the aspect ratio of the image to sometimes be lost.

#### Solution
Added properties to Image element.

1. "fitMode": "cover/fill/contain"
2. "verticalContentAlignment": "bottom/center/top"
3. "horizontalContentAlignment": "right/center/left" 

**In fitMode as fill, it takes full width and height and image aspect ratio is not maintained in this case.**
Screenshot 2025-06-12 at 10 33 08 AM

**In fitMode as cover, smaller dimension covers the full side and other side will get cropped.  Aspect ratio is maintained in this case.**
Screenshot 2025-06-12 at 10 33 30 AM

**In fitMode as contain, larger dimension covers the full side and other side will have some empty space.  Aspect ratio is maintained in this case. Red color is shown just for testing purpose. It will be removed before merging.**
Screenshot 2025-06-12 at 10 33 48 AM

**Demo video showing fitMode with verticalContentAlignment and horizontalContentAlignment.**

https://github.com/user-attachments/assets/6dc531d7-c31c-46c4-ab69-c033be81523d

#### More Fit Mode Changes
1. Use fitMode only if both width and height is provided in pixels.
2. If imageFitmode is contain then ImageStyle will always be Default.

#### Other changes

1. Downgraded ext.kotlin_version to '1.9.24'.
2. Added one more constructor to pass renderArgs to BaseActionElementRenderer listener.

### DIFF
--- a/source/android/adaptivecards/src/main/java/io/adaptivecards/renderer/readonly/ImageRenderer.java
+++ b/source/android/adaptivecards/src/main/java/io/adaptivecards/renderer/readonly/ImageRenderer.java
@@ -323,6 +323,7 @@ public class ImageRenderer extends BaseCardElementRenderer
                 break;
         }
 
+        // TODO :- Need to remove it before merging
         imageView.setBackgroundColor(Color.RED);
     }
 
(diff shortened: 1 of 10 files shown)


source/android/adaptivecards/src/main/java/io/adaptivecards/renderer/readonly/ImageRenderer.java
```diff
@@ -308,6 +310,86 @@ private int getBackgroundColorFromHexCode(String hexColorCode)
...
+
+            float viewWidth = imageView.getWidth();
+            float viewHeight = imageView.getHeight();
+
+            int dWidth = drawable.getIntrinsicWidth();
+            int dHeight = drawable.getIntrinsicHeight();
+
+            float scale = 1.0f;
+            switch (image.GetImageFitMode()) {
+                case Cover:
```
gauravkeshre, line 345: Same: Can we club duplicate logic using fall-through

gauravkeshre (approved):

meethjain (approved):

PR #393, opened 2025-06-14
fix: fix MicrosoftFluentUI/Tooltip_ios version

wood1986:
# Related Issue

N/A

# Description

I have an iOS app and it depends on AdaptiveCards and MicrosoftFluentUI. Here is my Podfile

image
When I do pod install, it returns this

image
Basically, for all projects depends on both AdaptiveCards and MicrosoftFluentUI are unable to complete pod install


mischreiber (approved):
Yep, this is the correct way that we expect intermediate podspecs to consume FluentUI. This maximizes flexibility for app consumers to pick and choose the versions they need for their apps.

PR #409, opened 2025-07-11
Fixing carousel view actions

P-Harika:
Carousel actions are not working.
Accessibility view is blocking the elements and making it inaccessible.


hggz (approved):
Will need to follow up on this to decide what to do with the invisible view

PR #434, opened 2025-08-25
iOS Input label changes

P-Harika:
Implemented below:
- TextBlock & RichTextBlock
	- use TextInput.getIsRequired
	- add red * as text only
- Input.Text
	- use TextInput.getLabel(element.getId())
	- announce as accessibility text, if required present append to text

### DIFF
--- a/source/ios/AdaptiveCards/AdaptiveCards/AdaptiveCards/ACRInputLabelView.mm
+++ b/source/ios/AdaptiveCards/AdaptiveCards/AdaptiveCards/ACRInputLabelView.mm
@@ -161,7 +161,7 @@
             NSString *accessibilityString = [NSString stringWithUTF8String:label.c_str()];
             if (inputBlck->GetIsRequired())
             {
-                accessibilityString = [accessibilityString stringByAppendingString: NSLocalizedString(@"Required", nil)];
+                accessibilityString = [accessibilityString stringByAppendingString: NSLocalizedString(@"\nRequired", nil)];
             }
             inputView.accessibilityLabel = accessibilityString;
         }
--- a/source/ios/AdaptiveCards/AdaptiveCards/AdaptiveCards/ACRInputLabelView.mm
+++ b/source/ios/AdaptiveCards/AdaptiveCards/AdaptiveCards/ACRInputLabelView.mm
@@ -12,6 +12,7 @@
 #import "ACRQuickReplyView.h"
 #import "UtiliOS.h"
 #import "ValueChangedAction.h"
+#import "TextInput.h"
 
 @implementation ACRInputLabelView
 
@@ -124,6 +125,13 @@
         } else if (!inputBlck->GetIsRequired()) {
             self.label.hidden = YES;
         }
+        // Remove - To be taken care in shared
+        // Desktop is also checking if id is matching with any of the labelFor values, if there's no match then they are showing original label
+        // Make that change in shared layer
+//        if (!inputBlck->GetLabelId().empty())
+//        {
+//            self.label.hidden = YES;
+//        }
 
         self.label.attributedText = attributedLabel;
         std::string errorMessage = inputBlck->GetErrorMessage();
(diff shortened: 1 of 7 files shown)


source/ios/AdaptiveCards/AdaptiveCards/AdaptiveCards/ACRInputLabelView.mm
```diff
@@ -147,6 +148,20 @@ - (instancetype)initInputLabelView:(ACRView *)rootView acoConfig:(ACOHostConfig
         self.inputView = inputView;
         self.label.isAccessibilityElement = NO;
         self.isAccessibilityElement = NO;
+        std::string label = TextInput().getLabel(inputBlck->GetId());
+        if (!label.empty())
```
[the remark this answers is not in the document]
P-Harika: Can you please help me understand this better ?
That's not a class variable. It's a local variable to check if there's any label associated with Id.
Are you refering to some other place ?

source/shared/cpp/ObjectModel/RichTextBlock.cpp
```diff
@@ -69,6 +76,26 @@ std::shared_ptr<BaseCardElement> RichTextBlockParser::Deserialize(ParseContext&
...
+    
+    if (!richTextBlock->m_labelFor.empty() && !richTextBlock->m_inlines.empty())
+    {
+        std::string label;
+        for (const auto &item: richTextBlock->m_inlines)
+        {
+            if (item->GetInlineType() == InlineElementType::TextRun)
+            {
+                auto textRun = std::static_pointer_cast<TextRun>(item);
+                label += textRun->GetText() + " ";
```
gauravkeshre, line 89: what is this empty space for?

source/ios/AdaptiveCards/AdaptiveCards/AdaptiveCards/ACRTextBlockRenderer.mm
```diff
@@ -113,6 +114,14 @@ - (UIView *)render:(UIView<ACRIContentHoldingView> *)viewGroup
             NSForegroundColorAttributeName : foregroundColor,
         }
                          range:NSMakeRange(0, content.length)];
+        
+        if (!txtBlck->GetLabelFor().empty() && TextInput().getIsRequired(txtBlck->GetLabelFor()))
+        {
+            RichTextElementProperties redStarProperties;
+            redStarProperties.SetTextColor(ForegroundColor::Attention);
+            NSAttributedString *redStar = initAttributedText(acoConfig, " *", redStarProperties, [viewGroup style]);
```
gauravkeshre, line 122: same here - explore NSMutableParagraphStyle and NSBaselineOffsetAttributeName instead of space

source/shared/cpp/ObjectModel/BaseInputElement.cpp
```diff
@@ -29,7 +30,8 @@ std::string BaseInputElement::GetLabel() const
 
 void BaseInputElement::SetLabel(const std::string label)
 {
-    m_label = label;
+    std::string inputLabel = TextInput().getLabel(GetId());
+    m_label = inputLabel.empty() ? label : "";
```
gauravkeshre, line 34: what does this line do?

gauravkeshre (approved):

source/shared/cpp/ObjectModel/RichTextBlock.cpp
P-Harika: Shared changes are from this PR:
https://github.com/microsoft/Teams-AdaptiveCards-Mobile/pull/430

source/ios/AdaptiveCards/AdaptiveCards/AdaptiveCards/ACRTextBlockRenderer.mm
P-Harika: Since we are using `RichTextElementProperties` I checked if we have such attribute to add paragraph style to it.
We don't have such property to add. The class is a shared class so it's better not to make changes in `RichTextElementProperties`.
Other approach is that we can avoid using `RichTextElementProperties` and use our attribues but the problem is attention text style color comes from hostConfig and the star should be colored as per the host config.
Keeping this as is.

source/ios/AdaptiveCards/AdaptiveCards/AdaptiveCards/ACRRichTextBlockRenderer.mm
```diff
@@ -199,6 +200,14 @@ - (UIView *)render:(UIView<ACRIContentHoldingView> *)viewGroup
                 [content appendAttributedString:textRunContent];
             }
         }
+        
+        if (!rTxtBlck->GetLabelFor().empty() && TextInput().getIsRequired(rTxtBlck->GetLabelFor()))
+        {
+            RichTextElementProperties redStarProperties;
+            redStarProperties.SetTextColor(ForegroundColor::Attention);
+            NSAttributedString *redStar = initAttributedText(acoConfig, " *", redStarProperties, [viewGroup style]);
```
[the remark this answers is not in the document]
P-Harika: Since we are using `RichTextElementProperties` I checked if we have such attribute to add paragraph style to it.
We don't have such property to add. The class is a shared class so it's better not to make changes in `RichTextElementProperties`.
Other approach is that we can avoid using `RichTextElementProperties` and use our attribues but the problem is attention text style color comes from hostConfig and the star should be colored as per the host config.
Keeping this as is.

gauravkeshre (approved):

PR #491, opened 2025-12-26
Exposed placeholder bool property in .h

P-Harika:
Exposed `isShowingPlaceholder` boolean
Reason: This boolean is used for validation purpose, if this is not properly set then the text field validation is failing


gauravkeshre:
We should not rely on view (that too placeholders) to perform validation. 
CAn we have another bool that indicates that the text field is empty?

PR #518, opened 2026-03-03
Fix: TalkBack does not announce Image role on ImageView

hggzm:
## Problem
TalkBack does not announce the "Image" role when focusing on ImageView elements in Adaptive Cards, making it difficult for visually impaired users to understand the content type.

## Solution
Added an AccessibilityDelegate to ImageView in ImageRenderer.java that sets roleDescription="Image" so TalkBack announces the element as an image.

## Changes
- **ImageRenderer.java**: Added AccessibilityDelegate with roleDescription "Image" on the ImageView

## Testing
- Verified with TalkBack on Android that images are now announced as "Image" role
- No impact on visual rendering


hggzm:
Superseded by #706.

This PR was raised from a fork, and fork-based PRs do not trigger this repository's CI — so its required checks could never run. #706 contains the **same change**, rebased onto current `main` and raised from an internal branch so CI can validate it.

Closing here in favour of #706 to avoid duplicate review.

PR #519, opened 2026-03-03
Fix: TalkBack announces both Link and Button roles for OpenUrl actions

hggzm:
## Problem
When TalkBack focuses on an OpenUrl action element, it announces both "Link" and "Button" roles, creating a confusing experience for screen reader users.

## Solution
Set roleDescription="Link" and clear the Button className on OpenUrl action elements so TalkBack only announces the "Link" role.

## Changes
- **ActionElementRenderer.java**: Added roleDescription "Link" and cleared Button className for OpenUrl actions

## Testing
- Verified with TalkBack on Android that OpenUrl actions announce only "Link" role
- Other action types (Submit, Execute) remain unaffected


hggzm:
Superseded by #707.

This PR was raised from a fork, and fork-based PRs do not trigger this repository's CI — so its required checks could never run. #707 contains the **same change**, rebased onto current `main` and raised from an internal branch so CI can validate it.

Closing here in favour of #707 to avoid duplicate review.

PR #520, opened 2026-03-03
Fix: TalkBack does not announce error messages on input validation failure

hggzm:
## Problem
When input validation fails, error messages appear visually but TalkBack does not announce them, leaving screen reader users unaware of validation errors.

## Solution
Set ACCESSIBILITY_LIVE_REGION_POLITE on the error text view and call announceForAccessibility() when an error message is shown, so TalkBack automatically announces validation errors.

## Changes
- **StretchableInputLayout.java**: Added live region and announceForAccessibility for error messages

## Testing
- Verified with TalkBack on Android that error messages are announced when validation fails
- No impact on visual rendering or non-accessibility behavior


hggzm:
Superseded by #708.

This PR was raised from a fork, and fork-based PRs do not trigger this repository's CI — so its required checks could never run. #708 contains the **same change**, rebased onto current `main` and raised from an internal branch so CI can validate it.

Closing here in favour of #708 to avoid duplicate review.

PR #521, opened 2026-03-03
Fix: TalkBack announces incorrect item count for dropdown (includes hidden placeholder)

hggzm:
## Problem
TalkBack announces an incorrect item count for dropdown/choice-set spinners because it counts the hidden placeholder item that is not visible to the user.

## Solution
Changed from m_items.size() to getCount() which returns the adapter count excluding the hidden placeholder item.

## Changes
- **ChoiceSetInputRenderer.java**: Changed item count from m_items.size() to getCount()

## Testing
- Verified with TalkBack on Android that the correct item count is announced
- Placeholder item is properly excluded from the count


hggzm:
Superseded by #709.

This PR was raised from a fork, and fork-based PRs do not trigger this repository's CI — so its required checks could never run. #709 contains the **same change**, rebased onto current `main` and raised from an internal branch so CI can validate it.

Closing here in favour of #709 to avoid duplicate review.

PR #522, opened 2026-03-03
Fix: TalkBack aggregates all radio button labels when focusing RadioGroup

hggzm:
## Problem
When TalkBack focuses on a RadioGroup in a ChoiceSet, it reads all child radio button labels concatenated together, creating a confusing and verbose announcement.

## Solution
Set IMPORTANT_FOR_ACCESSIBILITY_NO on the RadioGroup container so TalkBack skips the group and focuses directly on individual radio buttons.

## Changes
- **ChoiceSetInputRenderer.java**: Added IMPORTANT_FOR_ACCESSIBILITY_NO on RadioGroup

## Testing
- Verified with TalkBack on Android that individual radio buttons are focused separately
- Each radio button announces only its own label


hggzm:
Superseded by #710.

This PR was raised from a fork, and fork-based PRs do not trigger this repository's CI — so its required checks could never run. #710 contains the **same change**, rebased onto current `main` and raised from an internal branch so CI can validate it.

Closing here in favour of #710 to avoid duplicate review.

PR #523, opened 2026-03-03
Fix: ShowCard toggle announces selected/not selected instead of expanded/collapsed

hggzm:
## Problem
When toggling a ShowCard action, TalkBack announces "selected/not selected" instead of the more semantically correct "expanded/collapsed", which does not accurately convey the toggle behavior.

## Solution
Replaced setSelected() with announceForAccessibility() to explicitly announce "expanded" or "collapsed" state when toggling ShowCard visibility.

## Changes
- **BaseActionElementRenderer.java**: Replaced setSelected with announceForAccessibility for expanded/collapsed state

## Testing
- Verified with TalkBack on Android that ShowCard toggle announces "expanded" and "collapsed"
- Visual behavior remains unchanged

### DIFF
--- a/source/android/adaptivecards/src/main/java/io/adaptivecards/renderer/BaseActionElementRenderer.java
+++ b/source/android/adaptivecards/src/main/java/io/adaptivecards/renderer/BaseActionElementRenderer.java
@@ -332,7 +332,8 @@ public abstract class BaseActionElementRenderer implements IBaseActionElementRen
                 }
             }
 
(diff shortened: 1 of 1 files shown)


hggzm:
Superseded by #711.

This PR was raised from a fork, and fork-based PRs do not trigger this repository's CI — so its required checks could never run. #711 contains the **same change**, rebased onto current `main` and raised from an internal branch so CI can validate it.

Closing here in favour of #711 to avoid duplicate review.

PR #664, opened 2026-03-18
fix(ci): run assembleDebug instead of full build to avoid flaky packageRelease

hggzm:
## Problem

The Android PR CI (`teams-adaptivecards-android-pr`) intermittently fails with:

```
Execution failed for task ':mobile:packageRelease'.
> A failure occurred while executing com.android.build.gradle.tasks.PackageAndroidArtifact$IncrementalSplitterRunnable
```

This is a known Android Gradle Plugin issue where the incremental APK packager fails due to file locking race conditions on shared CI runners. The actual code compiles fine — `assembleDebug` succeeds, all unit tests pass — but the release packaging step fails randomly.

## Fix

Replace the `Gradle@3` ADO task with a direct script step that runs only:
- `assembleDebug` — verifies code compiles
- `testDebugUnitTest` — runs all unit tests

This skips `assembleRelease` / `packageRelease` entirely, which is unnecessary for PR validation. Release builds are handled by the separate release pipeline.

This matches the approach used for the iOS CI fix (PR #600) where we replaced `Xcode@5` with direct `xcodebuild` script steps.


GabrielMedAlv (approved):

hggz:
## Bump for review priority

This PR (along with #665) is currently blocking ~30 open accessibility PRs from passing CI due to a pre-existing flaky `packageRelease` step (#664) and a flaky popover UI test (#665). Both are GREEN and ready to merge.

Could a maintainer please prioritize review/merge of these two CI-fix PRs? Once they land, the rest of the a11y PR queue can be rebased and re-validated cleanly.

cc @karthikgopal36

hggzm:
This has been approved since 2026-03-20 and both pipelines are green; it is only sitting `BEHIND`. I have merged latest `main` so it is mergeable again.

Some data on what it is costing while it waits, from this week alone:

- **#729** (iOS-only, `ACRCompoundButtonRenderer.mm`) — iOS green, Android red on `packageRelease`
- **#730** (iOS-only, one character in `UtiliOS.mm`) — iOS green, Android red on `packageRelease`
- **#705** — currently red on the same step

Every one of those is an iOS-only change failing on an Android packaging step:

```
Unable to strip the following libraries, packaging them as they are:
  libadaptivecards-native-lib.so, libc++_shared.so
> Task :mobile:packageRelease FAILED
```

The failure has nothing to do with the changes under review, so each occurrence costs a retrigger and roughly 35 minutes of Android CI to get back to the same place. It also trains reviewers to ignore a red Android check, which is the more expensive problem — a genuine Android regression would look identical.

The change itself is one file and narrow: run `assembleDebug testDebugUnitTest lint` instead of the full `build`, which skips the release packaging step that does the stripping. Debug builds and unit tests still run, so PR coverage is unchanged.

@GabrielMedAlv you approved this in March — would you be able to merge, or point me at whoever owns `.pipelines/` now?

PR #675, opened 2026-04-04
[Release] AdaptiveCards iOS SDK v2.11.17

hggzm:
### Version bump
- iOS podspec: 2.11.7 -> 2.11.17

This is a version bump to include all accumulated fixes on main since v2.11.7.


hggzm:
closing this as superseded. 2.12.0 is going out in #719 and main is already past this point, so merging this later would drag the podspec backwards from 2.12.0 to 2.11.17.

nothing is lost by not merging it. 2.11.17 was already tagged and published to trunk out of band back in april, and the only commit unique to that tag versus main is the version bump itself (ce4096a3b), no functional content.

PR #680, opened 2026-04-16
[ios][fix] Reverted a change that was causing regression and failing ST

gauravkeshre:
# Related Issue

Elements marked visible with all children marked iinvisible were treated as invisible. 
This was causing ST failure.

Reverted the check in ACRContentStackView


P-Harika (approved):

hggzm (approved):
Approved. Targeted revert of the visibility check that was over-collapsing containers whose children toggle visibility at runtime. Restores prior behavior and unblocks the failing ST. LGTM.

PR #691, opened 2026-06-12
[iOS] Fix missing bullet markers in TextBlock markdown lists.

GabrielMedAlv:
# Explain defect

TextBlocks containing markdown unordered lists (e.g. `"- Item 1"`) render on iOS without the leading bullet marker — the text appears but the `•` is missing. The shared C++ markdown parser correctly emits `<ul><li>Item 1</li></ul>`, but on modern iOS `NSHTMLTextDocumentType` renders the bullet via `NSTextList` inside the paragraph style rather than as a literal character in the text. `ACRTextBlockRenderer` then overwrites every paragraph style with a fresh `NSMutableParagraphStyle` (only `alignment` is set) to honor the host-config alignment, which wipes out `NSTextList`, head-indent, and tab-stops — leaving the text without its bullet marker.

# Explain fix

In `buildIntermediateResultForText` (`UtiliOS.mm`), preprocess the markdown→HTML output before feeding it to `NSAttributedString`: strip` <ul>/</ul>`, replace `<li> `with `"• ",` and replace `</li>` with `<br>`. The bullet now lives in the text itself, so it survives any downstream paragraph-style changes and renders consistently across iOS versions. This mirrors the Android renderer's `RendererUtil.UlTagHandler`, which already injects literal `"\n• "` bullets into the spannable. Because the fix is in the shared `buildIntermediateResultForText`, it applies uniformly to `TextBlock`, `RichTextBlock`, and `FactSet`.

### DIFF
--- a/source/ios/AdaptiveCards/AdaptiveCards/AdaptiveCards/UtiliOS.mm
+++ b/source/ios/AdaptiveCards/AdaptiveCards/AdaptiveCards/UtiliOS.mm
@@ -750,6 +750,19 @@ void buildIntermediateResultForText(ACRView *rootView, ACOHostConfig *hostConfig
     }
     NSString *parsedString = (markDownParser->HasHtmlTags()) ? [NSString stringWithCString:markdownString.c_str() encoding:NSUTF8StringEncoding] : [NSString stringWithCString:markDownParser->GetRawText().c_str() encoding:NSUTF8StringEncoding];
 
+    // Convert <ul>/<li> to literal "• " bullets and <br> separators.
+    // Apple's NSHTMLTextDocumentType renders list bullets via NSTextList in the
+    // paragraph style, which is overwritten downstream in ACRTextBlockRenderer
+    // when alignment is applied — wiping out the bullet markers. Mirroring the
+    // Android renderer (RendererUtil.UlTagHandler) by injecting bullets as text
+    // makes the marker survive any paragraph-style changes.
+    if (markDownParser->HasHtmlTags() && [parsedString containsString:@"<li>"]) {
+        parsedString = [parsedString stringByReplacingOccurrencesOfString:@"<ul>" withString:@""];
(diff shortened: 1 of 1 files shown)


VikrantSingh01 (approved):

source/ios/AdaptiveCards/AdaptiveCards/AdaptiveCards/UtiliOS.mm
```diff
@@ -750,6 +750,19 @@ void buildIntermediateResultForText(ACRView *rootView, ACOHostConfig *hostConfig
...
     NSString *parsedString = (markDownParser->HasHtmlTags()) ? [NSString stringWithCString:markdownString.c_str() encoding:NSUTF8StringEncoding] : [NSString stringWithCString:markDownParser->GetRawText().c_str() encoding:NSUTF8StringEncoding];
 
+    // Convert <ul>/<li> to literal "• " bullets and <br> separators.
+    // Apple's NSHTMLTextDocumentType renders list bullets via NSTextList in the
+    // paragraph style, which is overwritten downstream in ACRTextBlockRenderer
+    // when alignment is applied — wiping out the bullet markers. Mirroring the
+    // Android renderer (RendererUtil.UlTagHandler) by injecting bullets as text
+    // makes the marker survive any paragraph-style changes.
+    if (markDownParser->HasHtmlTags() && [parsedString containsString:@"<li>"]) {
+        parsedString = [parsedString stringByReplacingOccurrencesOfString:@"<ul>" withString:@""];
```
JoshuaChingMSFT: Should we be using regex? If `<ul>` exists without a `</ul>`, it'll end up as an empty string.

GabrielMedAlv: You are right Josh. I changed using regex but, considering 2 regex were needed and could be slower, improved with the classic implementation to detect begin/end using a list, which is a single pass of the string.

JoshuaChingMSFT (approved):

PR #698, opened 2026-07-08
[Android ][iOS] Fix unsafe AdaptiveBase64 decode handling

karthikgopal36:
# Description

Fixes unsafe Base64 decoding in  AdaptiveBase64Util  by validating malformed padding/input before decoding and avoiding unchecked writes into a pre-sized output buffer.

Changes

• Decode into a temporary vector using bounded  push_back  writes.
• Reject malformed Base64 inputs such as excessive padding, invalid length, invalid characters, and padding in the middle.
• Preserve valid padded and unpadded decode behavior.
• Add regression tests for valid unpadded inputs and malformed padding cases.


hggzm:
same fix as the mainline PR, this is the production SDK copy. confirmed the diff is identical to what landed cross-platform: shared cpp AdaptiveBase64Util, and both ios (ACRView.mm -> AdaptiveBase64Util::Decode) and android (JNI to the same cpp) go through it, so this one change hardens the data-uri image decode on both.

validated it out of band on our teams adaptivecards fork: ported the same diff, it compiles clean in the ios + android SDK build, and the ios UI tests rendering data:image backgroundImage cards (including a malformed-base64 one) all pass without crashing. the CodeQL c-cpp analyze here is green too.

heads up the teams-adaptivecards-ios-pr red is a pod install failure (infra), not the change, might be worth a re-run.

hggzm (approved):
verified this hardens the shared cpp base64 decode used by both ios (ACRView.mm) and android (JNI). malformed data-uri base64 is now safely rejected instead of writing past the output buffer. validated the same diff out of band on our fork: compiles ios+android and ui tests with malformed base64 cards pass without crashing. ios pipeline red is a pod install infra failure, not the change.

MengliElmendorf137 (approved):

PR #695, opened 2026-06-30
fix(ios): expose interactive Icon (selectAction) to VoiceOver

hggzm:
## Summary
An **`Icon` with a `selectAction`** rendered no accessibility properties, so it was **not an accessibility element** \u2014 VoiceOver users could not find or activate it via swipe. `ACRIconRenderer` sets no accessibility properties on the icon/wrapping view even when the icon is interactive.

## Change
`ACRIconRenderer.mm` \u2014 when the icon has a `selectAction`, expose the wrapping view to VoiceOver:

```objc
if (acoSelectAction) {
    NSString *iconAccessibilityLabel = configureForAccessibilityLabel(acoSelectAction, nil);
    if (!iconAccessibilityLabel.length) {
        iconAccessibilityLabel = [NSString stringWithCString:icon->GetName().c_str() encoding:NSUTF8StringEncoding];
    }
    if (iconAccessibilityLabel.length) {
        wrappingView.isAccessibilityElement = YES;
        wrappingView.accessibilityLabel = iconAccessibilityLabel;
        wrappingView.accessibilityTraits = UIAccessibilityTraitButton;
    }
}
```

Reuses `configureForAccessibilityLabel` (the same helper the Image renderer uses for select-action labels). **Decorative icons (no `selectAction`) remain non-accessible** \u2014 no new noise in the a11y tree.

## Accessibility evidence (before \u2192 after)
VoiceOver-equivalent element-tree dump for the same card (interactive icons):

| | a11y tree |
|---|---|
| **Before** | 8 elements \u2014 **no icon/button elements at all** |
| **After** | 10 elements incl. **`button \"Airplane\"`** + **`button \"Click me!, Icons can handle actions\"`** |

**Before** \u2014 icons absent from the a11y tree:
before

**After** \u2014 box `[button] Airplane`: the icon is now a focusable, named button:
after

## \u2753 Open question
When the `selectAction` has no `title`/`tooltip`, the fix falls back to the icon **name** (e.g. `\"Airplane\"`) as the accessible label. That is far better than the control being unreachable, but an icon name is a developer-facing identifier and may not be the ideal spoken label. **Would maintainers prefer a different fallback (e.g. a generic localized \"Button\", or no name with just the button trait) when no title/tooltip is provided?** See my comment below.

PR #701, opened 2026-07-31
fix(ios): expose interactive Icon (selectAction) to VoiceOver

hggzm:
## Summary
An **`Icon` with a `selectAction`** rendered no accessibility properties, so it was **not an accessibility element** \u2014 VoiceOver users could not find or activate it via swipe. `ACRIconRenderer` sets no accessibility properties on the icon/wrapping view even when the icon is interactive.

## Change
`ACRIconRenderer.mm` \u2014 when the icon has a `selectAction`, expose the wrapping view to VoiceOver:

```objc
if (acoSelectAction) {
    NSString *iconAccessibilityLabel = configureForAccessibilityLabel(acoSelectAction, nil);
    if (!iconAccessibilityLabel.length) {
        iconAccessibilityLabel = [NSString stringWithCString:icon->GetName().c_str() encoding:NSUTF8StringEncoding];
    }
    if (iconAccessibilityLabel.length) {
        wrappingView.isAccessibilityElement = YES;
        wrappingView.accessibilityLabel = iconAccessibilityLabel;
        wrappingView.accessibilityTraits = UIAccessibilityTraitButton;
    }
}
```

Reuses `configureForAccessibilityLabel` (the same helper the Image renderer uses for select-action labels). **Decorative icons (no `selectAction`) remain non-accessible** \u2014 no new noise in the a11y tree.

VoiceOver-equivalent element-tree dump for the same card (interactive icons):

| | a11y tree |
|---|---|
| **Before** | 8 elements \u2014 **no icon/button elements at all** |
| **After** | 10 elements incl. **`button \"Airplane\"`** + **`button \"Click me!, Icons can handle actions\"`** |

**Before** \u2014 icons absent from the a11y tree:
before

**After** \u2014 box `[button] Airplane`: the icon is now a focusable, named button:
after

## \u2753 Open question
When the `selectAction` has no `title`/`tooltip`, the fix falls back to the icon **name** (e.g. `\"Airplane\"`) as the accessible label. That is far better than the control being unreachable, but an icon name is a developer-facing identifier and may not be the ideal spoken label. **Would maintainers prefer a different fallback (e.g. a generic localized \"Button\", or no name with just the button trait) when no title/tooltip is provided?** See my comment below.

---

> **Re-opened from an internal branch.** This PR is identical in content to #695, which was raised from a fork (`hggzm/Teams-AdaptiveCards-Mobile`).
> Fork-based PRs do not trigger the repository's CI, so #695 could not complete its checks despite being reviewed and approved.
> This PR carries the **same commit contents**, rebased onto the current `main`, from the internal branch `users/hggzm/clean/fix-icon-selectaction-a11y` (matching the convention used by previously merged PRs such as #672).
>
> Supersedes #695. Review history and approval for the identical change are on #695.

### DIFF
--- a/source/ios/AdaptiveCards/AdaptiveCards/AdaptiveCards/ACRIconRenderer.mm
+++ b/source/ios/AdaptiveCards/AdaptiveCards/AdaptiveCards/ACRIconRenderer.mm
@@ -69,6 +69,23 @@
     std::shared_ptr<BaseActionElement> selectAction = icon->GetSelectAction();
     ACOBaseActionElement *acoSelectAction = [ACOBaseActionElement getACOActionElementFromAdaptiveElement:selectAction];
     addSelectActionToView(acoConfig, acoSelectAction, rootView, wrappingView, viewGroup);
+
+    // An interactive icon (one with a selectAction) must be reachable and named for
+    // VoiceOver. Without this the icon is not an accessibility element, so screen
+    // reader users cannot find or activate it. Use the selectAction tooltip/title as
+    // the accessible name; fall back to the icon name. Decorative icons (no
+    // selectAction) are intentionally left non-accessible.
+    if (acoSelectAction) {
+        NSString *iconAccessibilityLabel = configureForAccessibilityLabel(acoSelectAction, nil);
+        if (!iconAccessibilityLabel.length) {
+            iconAccessibilityLabel = [NSString stringWithCString:icon->GetName().c_str() encoding:NSUTF8StringEncoding];
+        }
+        if (iconAccessibilityLabel.length) {
+            wrappingView.isAccessibilityElement = YES;
+            wrappingView.accessibilityLabel = iconAccessibilityLabel;
+            wrappingView.accessibilityTraits = UIAccessibilityTraitButton;
+        }
+    }
     
     // Configure visibility for the wrapping view so toggle visibility actions work correctly
     configVisibility(wrappingView, elem);
--- a/source/ios/AdaptiveCards/AdaptiveCards/AdaptiveCards/ACRIconRenderer.mm
+++ b/source/ios/AdaptiveCards/AdaptiveCards/AdaptiveCards/ACRIconRenderer.mm
@@ -72,18 +72,28 @@
 
     // An interactive icon (one with a selectAction) must be reachable and named for
     // VoiceOver. Without this the icon is not an accessibility element, so screen
-    // reader users cannot find or activate it. Use the selectAction tooltip/title as
-    // the accessible name; fall back to the icon name. Decorative icons (no
-    // selectAction) are intentionally left non-accessible.
+    // reader users cannot find or activate it. Decorative icons (no selectAction) are
+    // intentionally left non-accessible.
+    //
+    // Use the first meaningful name: the action's title, then its tooltip, then the
+    // icon name. Deliberately not configureForAccessibilityLabel(), which *joins* title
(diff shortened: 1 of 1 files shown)


VikrantSingh01 (approved):

hggzm:

GabrielMedAlv (approved):

hggzm:
**Pushed a correctness fix** for the `testPopoverRendering` failure:

```
testPopoverRendering, ((popoverIcon.exists && popoverIcon.isHittable) is true) failed
  - Button ', Click me to show a popover' should exist and be hittable
```

**This was a real defect in this change, not a flaky test.** It reproduced only once this change was combined with the others (the same test passed on #700, which did not include it).

**Cause.** The label was built with `configureForAccessibilityLabel()`, which *joins* the action's title and tooltip with `", "`:

```objc
if (action.title) { [labels addObject:action.title]; }   // empty, but non-nil
if (action.tooltip) { [labels addObject:action.tooltip]; }
return [labels componentsJoinedByString:@", "];          // -> ", Click me to show a popover"
```

For an `Action.Popover` with a tooltip and no title that yields **`", Click me to show a popover"`** — byte-identical to the label the action's own button already carries. The icon therefore became a *second* accessibility element with the same announced name:

- VoiceOver users hear the same name twice (the very class of defect #649 and #655 fix), and
- an exact-label lookup now matches two elements, so `.exists`/`.isHittable` cannot resolve.

**Fix.** Use the first meaningful name — title → tooltip → icon name — which is what this change always intended ("use the selectAction tooltip/title; fall back to the icon name"). The icon is now announced as **"Click me to show a popover"**: still reachable and named, no duplicate, and without the leading-comma pause.

Also changed `accessibilityTraits =` to `|=` so traits already applied by `setAccessibilityTrait()` (e.g. `UIAccessibilityTraitNotEnabled` for a disabled action) are preserved rather than clobbered.

**No test was modified** — the existing assertion is left exactly as-is and should now resolve to the single intended element.

hggzm:
Re-ran. The previous iOS failure was a flake, and there is direct cross-branch evidence for that:

```
build 11575 (this PR)  -> testPopoverRendering FAILED
build 11579 (PR #704)  -> testPopoverRendering PASSED (33.689s)
```

PR #704 carries a **byte-identical** `ACRIconRenderer.mm` patch (verified: same 28 added lines, `git diff main..<branch> -- ACRIconRenderer.mm` matches exactly). The same code passes the same test on the other branch, so the failure is not caused by this change.

For the record on why it *isn't* a real conflict — the button the test looks up is the **Image**, not the Icon:

- `ACRImageRenderer` sets `view.isAccessibilityElement = YES` and builds its label as `altText + configureForAccessibilityLabel(...)`. With empty `altText` and an action whose `title` is `""` and `tooltip` is `"Click me to show a popover"`, `configureForAccessibilityLabel` joins with `", "` and produces exactly `", Click me to show a popover"` — the string on line 833.
- `ACRIconRenderer` on `main` sets no accessibility properties at all. `addSelectActionToView` -> `setAccessibilityTrait` only assigns `userInteractionEnabled` and `accessibilityTraits`; it never sets `isAccessibilityElement` or a label. That is precisely the gap this PR closes.

This PR gives the Icon the label `"Click me to show a popover"` (no leading comma, via first-non-empty title/tooltip/name), which is deliberately **distinct** from the Image's `", Click me to show a popover"` — so no duplicate-name ambiguity is introduced.

GabrielMedAlv (approved):

PR #703, opened 2026-07-31
fix(ios): expose embedded citation buttons to VoiceOver / Full Keyboard Access / Voice Control

hggzm:
## Problem
Inline **citation pills** in Adaptive Cards are real `UIButton` views embedded via `NSTextAttachment` and positioned by `ACRViewAttachingTextViewBehavior`. Because the render surface is `ACRUILabel : UITextView` (whose `updateAccessibility` forces `isAccessibilityElement = YES`), the whole text view is exposed as a **single accessibility leaf** — so the embedded buttons are absent from **VoiceOver** traversal, **Full Keyboard Access**, and **Voice Control "Show Numbers"**. Direct touch works via the custom `hitTest:`; assistive technologies do not. Affects both **TextBlock** and **RichTextBlock** (both render via `ACRViewAttachingTextView`).

## Fix
- **`ACRViewAttachingTextViewBehavior`** — expose `orderedAttachments` (attached view + character range, in reading order) and an `attachmentsDidChangeHandler` fired after subviews are added/removed (`updateAttachedSubviews`) and repositioned (`layoutAttachedSubviews`).
- **`ACRViewAttachingTextView`** — when attachments exist, act as an **accessibility container**: `isAccessibilityElement` returns `NO` and `accessibilityElements` vends, in attributed-string order, virtual static-text elements (carrying `.header` / `.link` semantics with activation routing via `accessibilityActivationPoint`) interleaved with the real citation subviews (promoted to `.link`). Rebuilds on the behavior's change hook and posts `UIAccessibilityLayoutChangedNotification`. Clipped/truncated ranges are skipped (no offscreen elements). **When no attachments exist the behavior is unchanged**, preserving the existing single-element path for plain text blocks.

## Tests
`ACRCitationAccessibilityTests` (added to `AdaptiveCardsTextBlockTests.mm`):
- text view with a citation becomes an accessibility container (not a leaf) and exposes the pill;
- citation pill exposes `.link` semantics;
- reading order — leading text precedes the citation;
- plain-text no-regression — a block without citations does not engage the container.

## Validation
Built and tested on the iOS simulator (macOS 15 / Xcode). VoiceOver / Full Keyboard Access before-after evidence attached below.

## How verified
- Unit tests above (green on CI).
- Manual assistive-technology passes owed on-device before merge: VoiceOver, Full Keyboard Access, Voice Control "Show Numbers", plus Dynamic Type, RTL, streaming/text-replacement, headings, clipping, and direct-touch regression.

---

[...]

PR #699, opened 2026-07-21
fix(ios): expose embedded citation buttons to VoiceOver / Full Keyboard Access / Voice Control

hggzm:
## Problem
Inline **citation pills** in Adaptive Cards are real `UIButton` views embedded via `NSTextAttachment` and positioned by `ACRViewAttachingTextViewBehavior`. Because the render surface is `ACRUILabel : UITextView` (whose `updateAccessibility` forces `isAccessibilityElement = YES`), the whole text view is exposed as a **single accessibility leaf** — so the embedded buttons are absent from **VoiceOver** traversal, **Full Keyboard Access**, and **Voice Control "Show Numbers"**. Direct touch works via the custom `hitTest:`; assistive technologies do not. Affects both **TextBlock** and **RichTextBlock** (both render via `ACRViewAttachingTextView`).

## Fix
- **`ACRViewAttachingTextViewBehavior`** — expose `orderedAttachments` (attached view + character range, in reading order) and an `attachmentsDidChangeHandler` fired after subviews are added/removed (`updateAttachedSubviews`) and repositioned (`layoutAttachedSubviews`).
- **`ACRViewAttachingTextView`** — when attachments exist, act as an **accessibility container**: `isAccessibilityElement` returns `NO` and `accessibilityElements` vends, in attributed-string order, virtual static-text elements (carrying `.header` / `.link` semantics with activation routing via `accessibilityActivationPoint`) interleaved with the real citation subviews (promoted to `.link`). Rebuilds on the behavior's change hook and posts `UIAccessibilityLayoutChangedNotification`. Clipped/truncated ranges are skipped (no offscreen elements). **When no attachments exist the behavior is unchanged**, preserving the existing single-element path for plain text blocks.

## Tests
`ACRCitationAccessibilityTests` (added to `AdaptiveCardsTextBlockTests.mm`):
- text view with a citation becomes an accessibility container (not a leaf) and exposes the pill;
- citation pill exposes `.link` semantics;
- reading order — leading text precedes the citation;
- plain-text no-regression — a block without citations does not engage the container.

## Validation
Built and tested on the iOS simulator (macOS 15 / Xcode). VoiceOver / Full Keyboard Access before-after evidence attached below.

## How verified
- Unit tests above (green on CI).
- Manual assistive-technology passes owed on-device before merge: VoiceOver, Full Keyboard Access, Voice Control "Show Numbers", plus Dynamic Type, RTL, streaming/text-replacement, headings, clipping, and direct-touch regression.

PR #704, opened 2026-07-31
fix(ios): accessibility rollup - overflow button name, interactive Icon, card selectAction children, citation buttons

hggzm:
## Summary

Rollup of four already-reviewed iOS accessibility fixes, combined into a single PR so they can be validated and merged in one CI pass instead of four.

Each constituent change is **byte-identical** to its standalone PR (verified per-file), cherry-picked onto current `main`. They touch disjoint files, so there are no interactions between them.

| Fix | Standalone PR | File(s) |
|---|---|---|
| Give the action overflow (`…`) button a descriptive accessible name | #700 | `ACRActionOverflowRenderer.mm` |
| Expose interactive Icon (`selectAction`) to VoiceOver | #701 | `ACRIconRenderer.mm` |
| Keep interactive children reachable when a card has a `selectAction` | #702 | `ACRRenderer.mm` |
| Expose embedded citation buttons to VoiceOver / Full Keyboard Access / Voice Control | #703 | `ACRViewAttachingTextView.mm`, `ACRViewAttachingTextViewBehavior.h/.mm`, `AdaptiveCardsTextBlockTests.mm` |

Total: **7 files, +314 lines** (no deletions).

## Why a rollup

All four were reviewed and approved individually (on their original fork-based PRs #694, #695, #696, #699), but fork PRs do not trigger this repository's CI, so they could not complete their checks. They were re-raised from internal branches as #700–#703; this PR consolidates them so a single `teams-adaptivecards-ios-pr` / `teams-adaptivecards-android-pr` run covers all four.

## What each change does

**Overflow button name (#700)** — the overflow `…` button had no accessible name, so VoiceOver announced "dot dot dot". Gives it a localized descriptive name.

**Interactive Icon (#701)** — an `Icon` with a `selectAction` was not reachable by VoiceOver at all. Exposes it with the button trait and a label derived from `title` → `tooltip` → icon name.

**Card selectAction children (#702)** — when a card has a `selectAction`, its interactive children collapsed into a single unreachable element. Keeps children reachable and groups them for correct reading order.

**Citation buttons (#703)** — citation pills are real `UIButton`s embedded via `NSTextAttachment`, but the containing `ACRUILabel : UITextView` forced `isAccessibilityElement = YES`, presenting one leaf and hiding them from VoiceOver, Full Keyboard Access, and Voice Control "Show Numbers". Makes the text view an accessibility container that interleaves virtual static-text elements with the real citation subviews in reading order. Unchanged when no citations are present.

## Testing

`ACRCitationAccessibilityTests` (in `AdaptiveCardsTextBlockTests`) covers the citation container: container-vs-leaf, link semantics, reading order, and the plain-text no-regression contract. The other three are renderer-level accessibility attribute changes with before/after VoiceOver evidence on their standalone PRs.

## Merge note

If you prefer to merge this rollup, #700–#703 can be closed as included here. If you would rather take them individually, this PR can be closed instead — the standalone PRs remain valid and independent.

### DIFF
--- a/source/ios/AdaptiveCards/ADCIOSVisualizer/ADCIOSVisualizerUITests/ADCIOSVisualizerUITests.mm
+++ b/source/ios/AdaptiveCards/ADCIOSVisualizer/ADCIOSVisualizerUITests/ADCIOSVisualizerUITests.mm
@@ -495,7 +495,7 @@
     [textField typeText:@"Input inside popover\n"];
     
     // Click on overflow button
-    XCUIElement *overflowButton = [testApp.buttons elementMatchingPredicate:[NSPredicate predicateWithFormat:@"label CONTAINS[c] %@", @"..."]];
+    XCUIElement *overflowButton = [testApp.buttons elementMatchingPredicate:[NSPredicate predicateWithFormat:@"label CONTAINS[c] %@", @"More options"]];
     XCTAssertTrue(overflowButton.exists);
     [overflowButton tap];
     
@@ -554,7 +554,7 @@
     [textField typeText:@"Input inside popover\n"];
     
     // Click on overflow button
-    XCUIElement *overflowButton = [testApp.buttons elementMatchingPredicate:[NSPredicate predicateWithFormat:@"label CONTAINS[c] %@", @"..."]];
+    XCUIElement *overflowButton = [testApp.buttons elementMatchingPredicate:[NSPredicate predicateWithFormat:@"label CONTAINS[c] %@", @"More options"]];
     XCTAssertTrue(overflowButton.exists);
     [overflowButton tap];
     
@@ -731,7 +731,7 @@
     [textField typeText:@"1234\n"];
     
     // Click on overflow button
-    XCUIElement *overflowButton = [testApp.buttons elementMatchingPredicate:[NSPredicate predicateWithFormat:@"label CONTAINS[c] %@", @"..."]];
+    XCUIElement *overflowButton = [testApp.buttons elementMatchingPredicate:[NSPredicate predicateWithFormat:@"label CONTAINS[c] %@", @"More options"]];
     XCTAssertTrue(overflowButton.exists);
     [overflowButton tap];
     
@@ -797,7 +797,7 @@
     [textField typeText:@"1234\n"];
     
     // Click on overflow button
-    XCUIElement *overflowButton = [testApp.buttons elementMatchingPredicate:[NSPredicate predicateWithFormat:@"label CONTAINS[c] %@", @"..."]];
+    XCUIElement *overflowButton = [testApp.buttons elementMatchingPredicate:[NSPredicate predicateWithFormat:@"label CONTAINS[c] %@", @"More options"]];
     XCTAssertTrue(overflowButton.exists);
     [overflowButton tap];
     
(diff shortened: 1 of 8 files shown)


hggzm:
**Pushed a test fix** for the 4 `ADCIOSVisualizerUITests` failures in the last CI run:

```
testPopoverInput1Submission,           ((overflowButton.exists) is true) failed
testPopoverInput1SuccessfulSubmission, ((overflowButton.exists) is true) failed
testPopoverInput2Submission,           ((overflowButton.exists) is true) failed
testPopoverInput2SuccessfulSubmission, ((overflowButton.exists) is true) failed
```

**Cause (expected consequence of this change, not a flake).** Those tests locate the overflow control by its label:

```objc
[testApp.buttons elementMatchingPredicate:
    [NSPredicate predicateWithFormat:@"label CONTAINS[c] %@", @"..."]];
```

XCUITest surfaces `accessibilityLabel` as an element's `label`. This PR sets the overflow button's accessible name to the descriptive, localized "More options" (so VoiceOver stops reading "dot dot dot"), which replaces the literal `...` title — so the old predicate no longer matches.

**Fix.** Updated those 4 lookups to match the new accessible name (`@"More options"`). Test-only change; no production behaviour altered.

This is exactly the assertion the fix is meant to change: the button should no longer be identified by "...".

GabrielMedAlv (approved):

hggzm:
**Pushed a correctness fix** for the `testPopoverRendering` failure:

```
testPopoverRendering, ((popoverIcon.exists && popoverIcon.isHittable) is true) failed
  - Button ', Click me to show a popover' should exist and be hittable
```

**This was a real defect in this change, not a flaky test.** It reproduced only once this change was combined with the others (the same test passed on #700, which did not include it).

**Cause.** The label was built with `configureForAccessibilityLabel()`, which *joins* the action's title and tooltip with `", "`:

```objc
if (action.title) { [labels addObject:action.title]; }   // empty, but non-nil
if (action.tooltip) { [labels addObject:action.tooltip]; }
return [labels componentsJoinedByString:@", "];          // -> ", Click me to show a popover"
```

For an `Action.Popover` with a tooltip and no title that yields **`", Click me to show a popover"`** — byte-identical to the label the action's own button already carries. The icon therefore became a *second* accessibility element with the same announced name:

- VoiceOver users hear the same name twice (the very class of defect #649 and #655 fix), and
- an exact-label lookup now matches two elements, so `.exists`/`.isHittable` cannot resolve.

**Fix.** Use the first meaningful name — title → tooltip → icon name — which is what this change always intended ("use the selectAction tooltip/title; fall back to the icon name"). The icon is now announced as **"Click me to show a popover"**: still reachable and named, no duplicate, and without the leading-comma pause.

Also changed `accessibilityTraits =` to `|=` so traits already applied by `setAccessibilityTrait()` (e.g. `UIAccessibilityTraitNotEnabled` for a disabled action) are preserved rather than clobbered.

**No test was modified** — the existing assertion is left exactly as-is and should now resolve to the single intended element.

hggzm:
Re-ran. The previous iOS failure was the known keyboard-focus flake, not a regression:

```
testPopoverInput1SuccessfulSubmission
  Failed to synthesize event: Neither element nor any descendant has keyboard focus.
```

Worth noting from the same run: `testPopoverRendering` **passed** here (33.689s) while it failed on #701 — which carries a byte-identical `ACRIconRenderer.mm` patch. That cross-branch comparison is what confirms both popover failures are flakes rather than anything caused by these changes.

No source changes in the re-run.

hggzm:
closing since all sub branches from this rollup have been merged

PR #717, opened 2026-08-03
[Release] AdaptiveCards iOS SDK v2.11.18

hggzm:
## Summary

Version bump of the AdaptiveCards iOS pod to **2.11.18**.

Single-line change to `source/ios/tools/AdaptiveCards.podspec`. No source changes in this PR.

## Why 2.11.18 and not 2.11.10

CocoaPods resolves by SEMVER, not by publish date. Trunk currently holds an out-of-band
`2.11.17` (published 2026-04-04) that is higher than the current line `2.11.9`
(published 2026-07-14). Publishing `2.11.10` would sort BELOW `2.11.17`, so:

* trunk's "latest" would remain `2.11.17`
* `pod outdated` would keep reporting incorrectly
* any consumer on `~> 2.11` would never receive the fixes below

`2.11.18` restores a correct ordering. Neither `2.11.10` nor `2.11.18` exists on trunk today.

## What ships in 2.11.18

Everything merged to `main` since tag `iOS/adaptivecards-ios@2.11.9` (`8f36f5c27`).
Range `8f36f5c27..da1c696fa` is 10 files, +428 / -43.

### iOS accessibility fixes

| PR | File | Fix |
|----|------|-----|
| #700 | `ACRActionOverflowRenderer.mm` | overflow `...` button gets a localized accessible name |
| #701 | `ACRIconRenderer.mm` | interactive Icon (`selectAction`) exposed to VoiceOver, named by title/tooltip |
| #702 | `ACRRenderer.mm` | interactive children stay reachable when the card itself has a `selectAction` |
| #703 | `ACRViewAttachingTextView.mm`, `ACRViewAttachingTextViewBehavior.{h,mm}` | embedded citation buttons exposed to VoiceOver / Full Keyboard Access / Voice Control |

Tests: `AdaptiveCardsTextBlockTests.mm` (+90), `ADCIOSVisualizerUITests.mm` (+8/-4).

### Shared ObjectModel (C++) fix

| PR | File | Fix |
|----|------|-----|
| #698 | `shared/cpp/ObjectModel/AdaptiveBase64Util.cpp`, `Base64Test.cpp` | harden unsafe AdaptiveBase64 decode handling |

Note: #698 lives in the shared C++ ObjectModel, so it affects **both** iOS and Android.
The Android CI leg on this PR is therefore load-bearing and should not be dismissed as
infra without evidence.

## Public API

No public API breaks.

## Verification

[...]

PR #719, opened 2026-08-03
[iOS] Version upgrade to 2.12.0.

GabrielMedAlv:
Podspec file upgraded to 2.12.0.

### DIFF
--- a/source/ios/tools/AdaptiveCards.podspec
+++ b/source/ios/tools/AdaptiveCards.podspec
@@ -1,7 +1,7 @@
 Pod::Spec.new do |spec|
   spec.name             = 'AdaptiveCards'
 
-  spec.version          = '2.11.9'
+  spec.version          = '2.12.0'
 
   spec.license          = { :type => 'Adaptive Cards Binary EULA', :file => 'source/EULA-Non-Windows.txt' } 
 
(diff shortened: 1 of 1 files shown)


hggzm:
audited this against what i was about to ship as 2.11.18 and it supersedes mine cleanly, so im closing #717.

confirmed 2.12.0 picks up all of it: the four ios a11y fixes (#700 #701 #702 #703) plus the base64 hardening (#698), all ancestors of your head. also yours is more correct than mine was, i only bumped spec.version and missed the spec.source tag, you updated both.

on the red android leg, thats the known strip flake and not your change. build 12070 on your branch passed at 21:04, then 12076 on the merge ref failed at 21:24 with the same tree. failure is "Unable to strip the following libraries ... libadaptivecards-native-lib.so, libc++_shared.so" ending in :uitestapp:packageRelease FAILED. a podspec version string cant reach the gradle build at all. my 2.11.18 bump hit the identical failure. just needs a retrigger, and #664 is the real fix if it lands.

one thing worth checking before you publish: spec.source points at microsoft/AdaptiveCards-Mobile, so the iOS/adaptivecards-ios@2.12.0 tag has to exist in that mirror and not only here, otherwise pod install resolves nothing.

hggzm (approved):
lgtm, approving.

audited this against what i was about to ship as 2.11.18 and it is complete. all four ios a11y fixes (#700 #701 #702 #703) and the base64 hardening (#698) are ancestors of your head, and the release diff off 2.11.9 is the same 11 files i had.

also you updated both spec.version and the spec.source tag, which is the bit i got wrong in mine, so this is the correct shape.

android is the known strip flake, im retriggering that build now.

arocc (approved):

PR #723, opened 2026-08-05
ios: add thread safe text data

JoshuaChingMSFT:
# Description

Adds a thread-safe API for retrieving preprocessed text data from `ACRView`.

`ACRView` stores text preprocessing results in a shared mutable dictionary. Writes are serialized through `_serial_text_queue`, while callers retrieving the live dictionary through `getTextMap` cannot coordinate individual reads with background preprocessing.

This change adds `textDataForElementId:`. The accessor reads the requested entry on the same serial queue used by writers, allowing consumers to safely retrieve text data without changing the existing `getTextMap` contract or SDK renderer behavior.

The change is intentionally narrow and preserves existing API semantics.

# Sample Card

```json
{
  "type": "AdaptiveCard",
  "version": "1.5",
  "body": [
    {
      "type": "Container",
      "items": [
        {
          "type": "RichTextBlock",
          "inlines": [
            { "type": "TextRun", "text": "Rich text 01 with background preprocessing. " },
            { "type": "TextRun", "text": "Rich text 02 with background preprocessing. ", "weight": "Bolder" },
            { "type": "TextRun", "text": "Rich text 03 with background preprocessing. " },
            { "type": "TextRun", "text": "Rich text 04 with background preprocessing. ", "italic": true },
            { "type": "TextRun", "text": "Rich text 05 with background preprocessing. " },
            { "type": "TextRun", "text": "Rich text 06 with background preprocessing. ", "weight": "Bolder" },
            { "type": "TextRun", "text": "Rich text 07 with background preprocessing. " },
            { "type": "TextRun", "text": "Rich text 08 with background preprocessing. ", "italic": true },
            { "type": "TextRun", "text": "Rich text 09 with background preprocessing. " },
            { "type": "TextRun", "text": "Rich text 10 with background preprocessing. ", "weight": "Bolder" },
            { "type": "TextRun", "text": "Rich text 11 with background preprocessing. " },
            { "type": "TextRun", "text": "Rich text 12 with background preprocessing. ", "italic": true }
          ]
        },
        {
          "type": "FactSet",
          "facts": [
            { "title": "Fact 01", "value": "Value with **markdown** and https://example.com/01" },
            { "title": "Fact 02", "value": "Value with **markdown** and https://example.com/02" },
            { "title": "Fact 03", "value": "Value with **markdown** and https://example.com/03" },
            { "title": "Fact 04", "value": "Value with **markdown** and https://example.com/04" },
            { "title": "Fact 05", "value": "Value with **markdown** and https://example.com/05" },
            { "title": "Fact 06", "value": "Value with **markdown** and https://example.com/06" },
            { "title": "Fact 07", "value": "Value with **markdown** and https://example.com/07" },
            { "title": "Fact 08", "value": "Value with **markdown** and https://example.com/08" },
            { "title": "Fact 09", "value": "Value with **markdown** and https://example.com/09" },
            { "title": "Fact 10", "value": "Value with **markdown** and https://example.com/10" },
            { "title": "Fact 11", "value": "Value with **markdown** and https://example.com/11" },
            { "title": "Fact 12", "value": "Value with **markdown** and https://example.com/12" }
          ]
        },
        {
          "type": "Container",
          "items": [
            {
              "type": "TextBlock",
              "text": "**TextBlock 01** [link](https://example.com/a) with enough content to exercise markdown conversion.",
              "wrap": true
            },
            {
              "type": "Container",
              "items": [
                { "type": "TextBlock", "text": "**TextBlock 02** [link](https://example.com/b) inside a nested container.", "wrap": true },
                { "type": "TextBlock", "text": "**TextBlock 03** [link](https://example.com/c) inside a nested container.", "wrap": true },
                { "type": "TextBlock", "text": "**TextBlock 04** [link](https://example.com/d) inside a nested container.", "wrap": true },
                { "type": "TextBlock", "text": "**TextBlock 05** [link](https://example.com/e) inside a nested container.", "wrap": true },
                { "type": "TextBlock", "text": "**TextBlock 06** [link](https://example.com/f) inside a nested container.", "wrap": true },
                { "type": "TextBlock", "text": "**TextBlock 07** [link](https://example.com/g) inside a nested container.", "wrap": true },
                { "type": "TextBlock", "text": "**TextBlock 08** [link](https://example.com/h) inside a nested container.", "wrap": true }
              ]
            }
          ]
        }
      ]
    }
  ]
}
```

# How Verified

1. Added unit tests covering:
   - retrieval of stored text preprocessing data;
   - missing and nil element IDs;
   - 500 concurrent reads and writes with final-value assertions.
2. Ran `AdaptiveCardsTextBlockTests` successfully on an iPhone 16 Pro iOS 18.6 simulator.
3. Ran the concurrent accessor test with Thread Sanitizer enabled; no data race was reported.
4. Manually rendered the sample card in ADCIOSVisualizer and confirmed the synchronized accessor returned the expected data across repeated TextBlock renders.

### DIFF
--- a/source/ios/AdaptiveCards/AdaptiveCards/AdaptiveCards/ACRFactSetRenderer.mm
+++ b/source/ios/AdaptiveCards/AdaptiveCards/AdaptiveCards/ACRFactSetRenderer.mm
@@ -53,8 +53,7 @@
     NSMutableAttributedString *content = nil;
     if (rootView) {
         std::shared_ptr<FactSet> fctSet = std::dynamic_pointer_cast<FactSet>(element);
-        NSMutableDictionary *textMap = [rootView getTextMap];
-        NSDictionary *data = textMap[elementId];
+        NSDictionary *data = [rootView textDataForElementId:elementId];
         NSData *htmlData = data[@"html"];
         NSDictionary *options = data[@"options"];
         NSDictionary *descriptor = data[@"descriptor"];
@@ -130,7 +129,6 @@
     [factSetWrapperView addArrangedSubview:valueStack];
     [ACRSeparator renderSeparationWithFrame:CGRectMake(0, 0, factSetConfig.spacing, factSetConfig.spacing) superview:factSetWrapperView axis:UILayoutConstraintAxisHorizontal];
 
-    NSMutableDictionary *textMap = [rootView getTextMap];
     NSInteger nValidFacts = 0;
 
     NSMutableArray *accessibilityElements = [[NSMutableArray alloc] init];
@@ -138,7 +136,7 @@
     for (auto fact : factSet->GetFacts()) {
         NSString *title = [NSString stringWithCString:fact->GetTitle().c_str() encoding:NSUTF8StringEncoding];
         NSString *titleElemId = [key stringByAppendingString:[[NSNumber numberWithInt:rowFactId++] stringValue]];
-        if (![textMap objectForKey:titleElemId]) {
+        if (![rootView textDataForElementId:titleElemId]) {
             RichTextElementProperties titleTextProp{factSetConfig.title, fact->GetTitle(), fact->GetLanguage()};
             buildIntermediateResultForText(rootView, acoConfig, titleTextProp, titleElemId);
         }
@@ -163,7 +161,7 @@
         }
         NSString *value = [NSString stringWithCString:fact->GetValue().c_str() encoding:NSUTF8StringEncoding];
         NSString *valElemId = [key stringByAppendingString:[[NSNumber numberWithInt:rowFactId++] stringValue]];
-        if (![textMap objectForKey:valElemId]) {
+        if (![rootView textDataForElementId:valElemId]) {
             RichTextElementProperties valueTextProp{factSetConfig.value, fact->GetValue(), fact->GetLanguage()};
             buildIntermediateResultForText(rootView, acoConfig, valueTextProp, valElemId);
         }
--- a/source/ios/AdaptiveCards/AdaptiveCards/AdaptiveCards/ACRView.h
+++ b/source/ios/AdaptiveCards/AdaptiveCards/AdaptiveCards/ACRView.h
@@ -45,7 +45,7 @@
 
 - (NSMutableDictionary *)getTextMap;
 
-// Returns the current text preprocessing result using synchronized access.
+// Returns nil for a nil or missing element ID. Do not call from the internal text-processing queue.
 - (NSDictionary *)textDataForElementId:(NSString *)elementId;
 
 - (ACOAdaptiveCard *)card;
--- a/source/ios/AdaptiveCards/AdaptiveCards/AdaptiveCards/ACRView.h
+++ b/source/ios/AdaptiveCards/AdaptiveCards/AdaptiveCards/ACRView.h
@@ -45,6 +45,9 @@
 
 - (NSMutableDictionary *)getTextMap;
 
+// Returns the current text preprocessing result using synchronized access.
+- (NSDictionary *)textDataForElementId:(NSString *)elementId;
+
 - (ACOAdaptiveCard *)card;
 
 - (UIView *)render;
--- a/source/ios/AdaptiveCards/AdaptiveCards/AdaptiveCards/ACRView.mm
+++ b/source/ios/AdaptiveCards/AdaptiveCards/AdaptiveCards/ACRView.mm
@@ -667,7 +667,7 @@ typedef UIImage * (^ImageLoadBlock)(NSURL *url);
         return nil;
     }
 
-    __block NSDictionary *data;
+    __block NSDictionary *data = nil;
     dispatch_sync(_serial_text_queue, ^{
         data = self->_textMap[elementId];
     });
--- a/source/ios/AdaptiveCards/AdaptiveCards/AdaptiveCards/ACRView.mm
+++ b/source/ios/AdaptiveCards/AdaptiveCards/AdaptiveCards/ACRView.mm
@@ -661,6 +661,19 @@ typedef UIImage * (^ImageLoadBlock)(NSURL *url);
     return _textMap;
 }
 
+- (NSDictionary *)textDataForElementId:(NSString *)elementId
+{
+    if (elementId == nil) {
+        return nil;
+    }
+
+    __block NSDictionary *data;
+    dispatch_sync(_serial_text_queue, ^{
+        data = self->_textMap[elementId];
+    });
+    return data;
+}
+
 - (ACOAdaptiveCard *)card
 {
     return _adaptiveCard;
--- a/source/ios/AdaptiveCards/AdaptiveCards/AdaptiveCardsTests/AdaptiveCardsTests.mm
+++ b/source/ios/AdaptiveCards/AdaptiveCards/AdaptiveCardsTests/AdaptiveCardsTests.mm
@@ -8,15 +8,43 @@
 #import "ACOBaseCardElementPrivate.h"
 #import "ACRBaseCardElementRenderer.h"
 #import "ACRContentHoldingUIView.h"
+#import "ACRFactSetRenderer.h"
 #import "ACRInputLabelView.h"
 #import "ACRRegistration.h"
 #import "ACRTextView.h"
+#import "ACRViewPrivate.h"
+#import "Fact.h"
+#import "FactSet.h"
 #import "TextBlock.h"
 #import "TextInput.h"
 #import <UIKit/UIKit.h>
 #import <XCTest/XCTest.h>
 
 using namespace AdaptiveCards;
+
+@interface ACRTrackingTextMapView : ACRView
+
+@property NSUInteger liveTextMapAccessCount;
+@property NSUInteger enqueueCount;
+
+@end
+
+@implementation ACRTrackingTextMapView
+
+- (NSMutableDictionary *)getTextMap
+{
+    self.liveTextMapAccessCount++;
+    return [super getTextMap];
+}
+
+- (void)enqueueIntermediateTextProcessingResult:(NSDictionary *)data
+                                      elementId:(NSString *)elementId
+{
+    self.enqueueCount++;
+    [super enqueueIntermediateTextProcessingResult:data elementId:elementId];
+}
+
+@end
 
 @interface AdaptiveCardsTests : XCTestCase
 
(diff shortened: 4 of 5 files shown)


hggzm (changes requested):
Nice, focused change — and thanks for routing the read through the *same* serial queue the writer uses rather than bolting on a second lock. The accessor itself is correct; my concerns are about scope and the test, so I'd like one more pass before signing off.

## What looks right

- `dispatch_sync(_serial_text_queue, …)` matches the writer in `enqueueIntermediateTextProcessingResult:` (`ACRView.mm:580`), so reads and writes are genuinely serialized against each other.
- The `nil` guard is load-bearing — subscripting `NSMutableDictionary` with a nil key raises, so `textDataForElementId:nil` would otherwise crash rather than return nil.
- `initWithFrame:` does create both `_textMap` and `_serial_text_queue` (`ACRView.mm:83-93`), so the new tests aren't standing on a partially-initialized object.
- Method naming follows ObjC convention, unlike the legacy `getTextMap`.

## 1. The race is still reachable after this change

The accessor currently has no callers, and the four real readers still take the raw dictionary and subscript it with no synchronization:

| Reader | Line |
|---|---|
| `ACRFactSetRenderer.mm` | 56, 133 |
| `ACRRichTextBlockRenderer.mm` | 81 |
| `ACRTextBlockRenderer.mm` | 75 (existence check **and** the following read) |

Writers run on the concurrent global queue — `processTextConcurrently:` is invoked at `ACRView.mm:248, 260, 274, 279` and dispatches via `dispatch_group_async(_async_tasks_group, _global_queue, …)` at `ACRView.mm:571`. Readers run on the render thread.

The one barrier that would make this safe, `waitForAsyncTasksToFinish` (`ACRView.mm:189`), has **no callers inside the SDK** — I grepped every `.mm`. So the SDK never establishes happens-before on its own; it's left entirely to the host app. That leaves a real read-during-write on a plain `NSMutableDictionary`, where a resize/rehash mid-read can return garbage or crash.

So the PR adds the safe door but leaves everyone walking through the unsafe one. Either resolution works for me:

1. **Migrate the four readers in this PR** (three are one-liners — see below), or
2. **Keep the PR narrow**, but state in the description that this is for host-app consumption and that the internal readers remain unsynchronized, and file a follow-up issue.

I only ask that we pick one explicitly rather than leave it implicit, because as it stands the title "add thread safe text data" reads as though the SDK is now thread-safe, and it isn't.

Related: `getTextMap` still vends the **mutable backing store**, so the unsafe path stays first-class next to the safe one. Worth deprecating it or returning `[_textMap copy]`.

## 2. The concurrency test doesn't demonstrate safety

`testTextMapSupportsConcurrentReadsAndWrites` would pass identically with or without this change:

- the reads are `(void)`-discarded, so no value is ever validated under contention;
- writes were already serialized before this PR;
- only 2 of 500 keys are asserted, and only after the group has fully drained;
- every key is written exactly once, so the dangerous case — reading a key while that same key is being overwritten — never occurs.

Concretely, I'd suggest:

```objc
// Validate the value under contention instead of discarding it,
// and bound the fan-out (500 blocking dispatch_syncs from a global
// concurrent queue can over-commit the thread pool in CI).
dispatch_apply(500, DISPATCH_APPLY_AUTO, ^(size_t i) {
    NSString *key = [NSString stringWithFormat:@"text-%zu", i];
    [rootView enqueueIntermediateTextProcessingResult:@{@"nonhtml" : key} elementId:key];
    XCTAssertEqualObjects([rootView textDataForElementId:key][@"nonhtml"], key);
});
```

plus a case that repeatedly overwrites one hot key from several threads while reading it, since that's the shape that actually corrupts an `NSMutableDictionary`.

One caveat worth stating plainly: **none of this proves the absence of a data race.** Only a sanitizer can do that, and I checked — there's no TSan-enabled scheme or sanitizer build setting anywhere in `source/ios` today, so that's new CI infrastructure rather than something to bolt onto this PR. I'd explicitly *not* recommend adding a test that races through `getTextMap` without TSan; it would just be an intermittently crashing test. Worth a separate infra item if we want real coverage here.

## 3. Minor

- Now that this is public header API, `dispatch_sync` on a serial queue deadlocks if ever called from that queue. Unreachable today, but a one-line header caveat would keep it that way.
- `__block NSDictionary *data;` is uninitialized. ARC zero-initializes strong locals so it's safe in practice, but `= nil` states the intent.
- `ACRView.h` has no `NS_ASSUME_NONNULL`, and the test deliberately passes nil. Annotating just this method would likely trip `-Wnullability-completeness` for the whole file, so I'd document nil-tolerance in the comment rather than annotate piecemeal.

## Accessibility impact — the reason I care about item 1

`_textMap` backs TextBlock, RichTextBlock and FactSet content, which is exactly what VoiceOver reads aloud. A torn or missed read doesn't just render wrong text — it can produce an element with **no text and therefore no accessible name**, which is effectively invisible to a screen-reader user. That's the strongest argument for migrating the readers rather than leaving the safe accessor unused.

---

## Suggested follow-up

Happy to file this as an issue if you'd rather keep this PR narrow.

**Migrate the readers.** Three are direct swaps, e.g. `ACRFactSetRenderer.mm:55-56`:

```diff
-        NSMutableDictionary *textMap = [rootView getTextMap];
-        NSDictionary *data = textMap[elementId];
+        NSDictionary *data = [rootView textDataForElementId:elementId];
```

`ACRTextBlockRenderer.mm` and `ACRRichTextBlockRenderer.mm` need a little more care — they read twice (existence check, then `buildIntermediateResultForText`, then the real read), so both reads should go through the accessor.

**Close the door.** Deprecate `getTextMap`, or have it return `[_textMap copy]` so callers can't hold the live mutable store.

**Optional infra.** Add a TSan-enabled test scheme so the race tests can actually assert what they claim.

JoshuaChingMSFT:
@hggzm Strengthened the concurrency tests and migrated the ACRFactSetRenderer reads to the thread-safe accessor. Keeping the remaining renderers unchanged is intentional so this PR stays limited to the confirmed crash sites and minimizes regression risk in unaffected areas.

GabrielMedAlv (approved):

hggzm (approved):
This addresses everything I raised — thanks for the thorough pass.

The `ACRTrackingTextMapView` guard is a better idea than what I suggested: asserting `liveTextMapAccessCount == 0` means the FactSet path can't silently regress back to a raw map read later. That's the part I'd have missed.

Your scoping rationale is sound and I'm happy with it. Migrating the confirmed crash site while leaving `ACRTextBlockRenderer` and `ACRRichTextBlockRenderer` alone is the lower-risk call, and the new accessor makes the remainder mechanical whenever we choose to do it. I've filed the follow-ups so nothing gets lost:

- #725 — migrate the two remaining readers and retire `getTextMap`
- #726 — add a TSan-enabled test scheme

On #726, for the record: the new concurrency tests are worth having, but I checked and there's no sanitizer-enabled scheme anywhere in `source/ios` today. So they validate that values read back under contention are correct — a real property — but they can't prove the *absence* of a race. That needs a sanitizer, which is infrastructure work rather than anything that belongs in this PR.

## Two non-blocking notes

**1. The description is now stale.** It still says the change is "without changing the existing `getTextMap` contract or SDK renderer behavior", but `ACRFactSetRenderer`'s read path did change. The rendered output is identical — it's a pure swap of data-access mechanism — but the sentence reads as though no renderer was touched. Worth a quick edit so the history is accurate.

**2. Per-fact queue round-trips — slightly more real than I first framed it.** FactSet now performs roughly four `dispatch_sync` hops per fact (two existence checks plus two `buildLabel` reads) on the render thread, where it previously did one `getTextMap` followed by plain dictionary lookups.

For the 2-row FactSets in the test cards that's noise. But real Teams incident and approval cards commonly run 8–15 rows, which is 30–60 serial round-trips during layout. Still very likely invisible — each write block is a single dictionary assignment, so contention should be minimal — but it's a bit more than "flag it if it shows up in profiles later". A single batched read per render pass would be cheap insurance. I've captured the shape of it in #725 rather than asking for it here, since correctness clearly comes first and the snapshot approach has a wrinkle (`buildIntermediateResultForText` writes *during* the render pass, so an up-front snapshot goes stale for elements built on demand).

**Minor:** the hardcoded `@"*0"` / `@"*1"` element IDs couple the tests to the renderer's internal ID format. Deriving them the way the renderer does would make a future ID change fail loudly rather than confusingly.

## One piece of context you couldn't have had

There is a **second, independent FactSet-visibility bug** in flight: #724, which stops `ACRContentStackView` collapsing containers that still have visible content. Zero file overlap with this PR (I checked), and neither fixes the other — yours is a data race producing crashes and torn reads, that one is an ordering bug where `hasVisibleContent` is queried before `_visibleViews` is populated, producing a collapsed container.

Flagging it so the two don't get conflated during triage, and so nobody assumes this PR resolves the "FactSet renders blank" reports — it won't. Both need to land for FactSet to be fully correct.

---

Approving. Note the ADO pipelines were still queued when I submitted, so this is contingent on `teams-adaptivecards-ios-pr` and `teams-adaptivecards-android-pr` going green — the five new tests haven't executed yet. The merge gate covers that, but shout if anything falls over and I'll take another look.

hggzm:
Following up on my approval caveat — the iOS pipeline came back red, but **not because of anything in this PR**. Worth spelling out, because the interesting part isn't the failure itself.

## The failure is an unrelated UI flake

```
✗ testDynamicTypeaheadSearchFromChoiceset
  Failed to scroll to visible (by AX action) Button,
  {{5.3, 46.4}, {123.3, 9.3}}, label: 'Delete All Cards',
  error: Error kAXErrorCannotComplete
```

That's a scroll/hittability failure in the sample-app UI suite (`ADCIOSVisualizerUITests`). Nothing in this PR touches typeahead, choicesets, or that button.

## The part that matters: your new tests still haven't run

Looking at the task breakdown for build 12557:

| Task | Scheme | Result |
|---|---|---|
| Sample App Unit Test | `ADCIOSVisualizer` | **failed** (the flake above) |
| Framework Unit Test | `AdaptiveCards` | **skipped** |

All five new tests — `testFactSetRendererUsesSynchronizedTextDataAccessor`, `testFactSetRendererUsesExistingPreprocessedTextData`, `testFactSetRendererSupportsConcurrentTextMapWrites`, `testTextMapSupportsConcurrentReadsAndWrites`, `testTextMapSupportsConcurrentHotKeyOverwrites` — live in the `AdaptiveCards` scheme, which is the **Framework Unit Test** step. That step has no `condition:` in `.pipelines/templates/ios-ci-template.yml`, so it defaults to `condition: succeeded()` and is skipped the moment the preceding UI-test step fails.

I confirmed this by searching the build log directly: none of the five test names appear anywhere in it. The only bundles that executed were `ADCIOSVisualizerTests` (65 tests, 0 failures) and `ADCIOSVisualizerUITests` (19 tests, 1 failure).

So the situation is that a flaky sample-app UI test prevented the SDK unit tests from running at all — including the ones written specifically to validate this change.

## What I've done

I've re-queued build 12557. **We need it to run to completion** before we can say the new tests actually pass — right now they remain unexecuted, so the `@"*0"` / `@"*1"` element-ID assumptions in the FactSet tests are still unverified.

My approval stands on the code, which I'm happy with. Just flagging that the green/red signal on this PR currently tells us nothing about the new tests either way.

## Broader issue

This isn't specific to your PR. Given the iOS UI suite's flake history in this repo — `testPopoverRendering`, `testPopoverInput1SuccessfulSubmission`, simulator-launch failures, and now this one — the Framework Unit Test step is being skipped fairly regularly, which means iOS SDK unit-test coverage is quietly lower than the check marks suggest.

It's a one-line fix (`condition: succeededOrFailed()`). I'll open a separate PR for it and link it here so it doesn't ride along with your change.

hggzm:
Opened #727 for the pipeline gap described above — `condition: succeededOrFailed()` on the Framework Unit Test step, so the SDK's unit tests always run instead of being skipped whenever the sample-app UI suite flakes.

Kept it as a separate PR so it doesn't ride along with your change. Once it merges, a red UI test will no longer hide whether the framework tests pass — which is exactly the situation this PR is stuck in right now.

The re-run of build 12557 is queued; I'll report back once the Framework Unit Test step actually executes and we can see how your five tests do.

hggzm:
Closing the loop on my approval caveat — the re-run of build `12557` came back green, the **Framework Unit Test** step executed this time, and all seven tests passed:

```
✓ testFactSetRendererUsesSynchronizedTextDataAccessor   (0.035s)
✓ testFactSetRendererUsesExistingPreprocessedTextData   (0.049s)
✓ testFactSetRendererSupportsConcurrentTextMapWrites    (2.964s)
✓ testTextMapSupportsConcurrentReadsAndWrites           (0.096s)
✓ testTextMapSupportsConcurrentHotKeyOverwrites         (0.008s)
✓ testTextDataForElementIdReturnsStoredData             (0.009s)
✓ testTextDataForElementIdReturnsNilForMissingOrNilKey  (0.006s)

Executed 367 tests, with 27 skipped and 0 failures
```

That also settles the one thing I couldn't verify statically: the `@"*0"` / `@"*1"` element IDs match what the renderer actually generates.

Thanks for the thorough iteration on this — the `liveTextMapAccessCount == 0` guard in particular is the piece that'll keep the FactSet path from quietly regressing.

PR #724, opened 2026-08-05
fix(iOS): stop collapsing containers that still have visible content

hggzm:
### The bug

`ACRContentStackView.updateLayoutAndVisibilityOfRenderedView:` collapses a child
container when it reports no visible content:

```objc
if (!acoElem.element->GetIsVisible()) {
    [self registerInvisibleView:renderedView];
} else if ([renderedView isKindOfClass:[ACRContentStackView class]] &&
           !((ACRContentStackView *)renderedView).hasVisibleContent) {
    [self registerInvisibleView:renderedView];
}
```

`hasVisibleContent` returns `_visibilityManager.hasVisibleViews`, i.e. `_visibleViews.count > 0`.
`_visibleViews` is populated **only** in `applyVisibilityToSubviews`, which by design runs
*after* every child has rendered.

So at the moment this check executes, a freshly rendered child `ACRContentStackView` has an
empty `_visibleViews` and **always** reports no visible content. Every nested container is
therefore registered invisible and collapsed, together with its content. It is deterministic,
not a race.

This is precisely the hazard `applyVisibilityToSubviews` already documents:

> this method applies visibility to subviews once all of them are rendered and become part of
> content stack view. applying visibility as each subview is rendered has known side effects.

### Impact

Introduced by 02d30e544 ("[iOS] Collapse ColumnSet/Container when all children are invisible"),
present since 2.11.8. Found while adopting 2.12.0 in Teams iOS: **17 reference cards lost
height with content missing**. A `Container` holding a `FactSet` rendered blank —
1050x288 collapsed to 1050x72. Another lost its entire FactSet (Category / Severity /
Location / Assigned to) while keeping only the header and button. Matching UI test failures:
`Work day app UI element missing` and `Stock update card mismatched!`.

Any card with a nested Container or ColumnSet is affected, which is most real-world cards.

### The change

Removes the inline check. This restores correct rendering and is the same resolution Teams
reached independently during the 2.11.8 adoption, so the behaviour is well understood.

The collapse feature itself is worth having — it just cannot be evaluated during rendering.
The right home is `applyVisibilityToSubviews`, where `_visibleViews` is actually populated.
I deliberately did not attempt that re-land here: I have no macOS environment to validate it,
and shipping an unverified ordering change in this area is how the original bug happened.

### Tests

Adds `ACRContainerCollapseTests` covering the shapes the collapse walked:

- `testContainerWithFactSetIsNotCollapsed` — the exact downstream failure
- `testNestedContainerWithVisibleChildIsNotCollapsed`
- `testColumnSetWithVisibleColumnIsNotCollapsed`
- `testContainerWithOnlyInvisibleChildrenRendersNoText` — keeps the genuinely-empty case
  honest, so a future re-land has a target to satisfy instead of silently regressing this file

There was no existing coverage for this: `hasVisibleContent` and `registerInvisibleView`
appear in no test file, and the only FactSet test is Android object-model. That gap is why
the same defect had to be caught twice by a downstream consumer's visual diff.

### Note for reviewers

`project.pbxproj` is edited to register the new test file (the project does not use
file-system-synchronized groups). That edit was made without Xcode, mirroring the exact
4-reference pattern of an existing test; brace balance verified. CI compiles the project, so
please treat a build failure there as my error rather than a flake.

### DIFF
--- a/source/ios/AdaptiveCards/AdaptiveCards/AdaptiveCards.xcodeproj/project.pbxproj
+++ b/source/ios/AdaptiveCards/AdaptiveCards/AdaptiveCards.xcodeproj/project.pbxproj
@@ -31,6 +31,7 @@
 		244049E12EC24097000A06DF /* ACRIImageResolver.h in Headers */ = {isa = PBXBuildFile; fileRef = 244049E02EC2408D000A06DF /* ACRIImageResolver.h */; settings = {ATTRIBUTES = (Public, ); }; };
 		244915622EBCE3BD0040DF94 /* ACOReferencePrivate.h in Headers */ = {isa = PBXBuildFile; fileRef = 244915612EBCE3BD0040DF94 /* ACOReferencePrivate.h */; settings = {ATTRIBUTES = (Public, ); }; };
 		246485662E3CEA0C0085F3BD /* ACRPopoverTargetTests.mm in Sources */ = {isa = PBXBuildFile; fileRef = 246485652E3CEA0C0085F3BD /* ACRPopoverTargetTests.mm */; };
+		724D78627BA31B464C0B33D0 /* ACRContainerCollapseTests.mm in Sources */ = {isa = PBXBuildFile; fileRef = 2086D6F3FBCB3C480E4E59BD /* ACRContainerCollapseTests.mm */; };
 		248924C62EBB617600F76802 /* CitationRun.cpp in Sources */ = {isa = PBXBuildFile; fileRef = 248924C32EBB617600F76802 /* CitationRun.cpp */; };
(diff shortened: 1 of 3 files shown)


P-Harika (approved):

arocc (approved):

GabrielMedAlv (approved):

hggz:
Rolled up into #739 together with the other fix, so we only have to fight the currently repo-wide `teams-adaptivecards-ios-pr` failure once instead of twice. This PR stays open and valid - close whichever of the two you prefer not to take.

hggz:
Included in #739, which is now merged (b06c114bdc04). Closing - the rollup carries this exact change byte-identical, plus the testPopoverRendering determinism fix that unblocked CI.

hggz:
Included in #739, which is merged. Closing this one - the rollup carries this exact change byte-identical, plus the testPopoverRendering determinism fix that unblocked CI.

PR #731, opened 2026-08-18
fix(ios): announce rating stars as buttons and repair the read-only rating label

hggzm:
## Summary

Two accessibility defects in `ACRRatingView`:

1. **Interactive rating stars are announced as images.** They are accessibility elements with good labels (`"Rate 3 Star"`) but carry no trait, so VoiceOver reads *"Rate 3 Star, image"* and gives no indication the control can be activated.
2. **The read-only rating label drops its count.** A rating that has a count announces `"Rating 3.2,"` — count omitted, dangling comma — while a rating with no count announces the meaningless `"Count 0"`.

## Root cause

**Stars** — `setupViewForInput` sets `isAccessibilityElement` and a label but no trait:

```objc
starImageView.isAccessibilityElement = YES;
starImageView.accessibilityLabel = [NSString stringWithFormat:@"Rate %d Star", (int)i+1];
UITapGestureRecognizer *tapGesture = ...
```

**Label** — the branches in `setupReadOnlyView` are inverted:

```objc
if (_count > 0)  { @"Rating %.1f,"          }   // has a count -> count dropped
else             { @"Rating %.1f, Count %d" }   // no count -> announces "Count 0"
```

The trailing comma in the first format is the giveaway: it was written to be followed by the count, which that branch never appends.

## Change

`ACRRatingView.mm` — OR in `UIAccessibilityTraitButton` for each tappable star, and swap the count branches so a rating with a count announces it and one without simply omits it.

VoiceOver element scan of `samples/v1.5/Scenarios/RatingInput.json` through the real UIAccessibility API (`XCUIElement`), against an **otherwise identical control build** — same harness, only this file differs:

| | a11y tree |
|---|---|
| **before** | 17 elements — `{ image: 10, text: 4, textView: 2, button: 1 }` |
| **after** | 17 elements — `{ button: 11, text: 4, textView: 2 }` |

Element level:

| | before | after |
|---|---|---|
| Rate 1–5 Star (×10) | `role=image` | `role=button` |
| read-only rating 1 | `"Rating 3.2,"` | `"Rating 3.2, Count 10"` |
| read-only rating 2 | `"Rating 3.2,"` | `"Rating 3.2, Count 150"` |
| read-only rating 3 | `"Rating 3.2,"` | `"Rating 3.2, Count 1500"` |

The three read-only ratings on this card have **different** counts (10, 150, 1500) and every one was announcing identically. Element count is unchanged at 17, so nothing was added or removed — these are role and label corrections only.

### Annotated overlay

Rating input accessibility overlay

Raw per-element dump:
https://hggzm.github.io/Teams-AdaptiveCards-Mobile/a11ymas_5535831_rating_input_elements.json

## Scope

One file. No layout or behaviour change; tap handling is untouched.

Work item: **AB#5535831** (A11yMAS, A11ySev2)

### DIFF
--- a/source/ios/AdaptiveCards/ADCIOSVisualizer/ADCIOSVisualizerUITests/ADCIOSVisualizerUITests.mm
+++ b/source/ios/AdaptiveCards/ADCIOSVisualizer/ADCIOSVisualizerUITests/ADCIOSVisualizerUITests.mm
@@ -599,7 +599,10 @@
 
     // Adjust this to select a rating (e.g., tap the 4th star for rating=4)
     // Tap the 4th star for rating=4
-    XCUIElement *fourthStar = [testApp.images elementMatchingPredicate:[NSPredicate predicateWithFormat:@"label == %@", @"Rate 4 Star"]];
+    // The star now carries UIAccessibilityTraitButton, so it resolves as a button rather
+    // than an image. That role change is the point of the fix - VoiceOver previously
+    // announced "Rate 4 Star, image" with no indication the control was actuatable.
+    XCUIElement *fourthStar = [testApp.buttons elementMatchingPredicate:[NSPredicate predicateWithFormat:@"label == %@", @"Rate 4 Star"]];
     XCTAssertTrue(fourthStar.exists && fourthStar.isHittable, @"The 4th star should exist and be hittable");
     [fourthStar tap];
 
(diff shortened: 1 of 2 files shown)


VikrantSingh01 (commented):
Pls check the failing check.

gauravkeshre (approved):

PR #734, opened 2026-08-25
fix(ios): restore VoiceOver focus after the overflow menu is cancelled

hggzm:
## Problem

`ACROverflowTarget` presents the overflow action sheet with `handler:nil` on its Cancel action, and posts no accessibility notification when the sheet is dismissed. Nothing tells iOS where VoiceOver focus should land, so it falls back to the platform default — the top of the card — instead of the control the user opened the menu from.

`ACRShowCardTarget` already handles the equivalent case by posting `UIAccessibilityLayoutChangedNotification` with an explicit anchor view. This applies that same established pattern to the overflow menu.

## Change

- `ACROverflowTarget` gains a weak `accessibilityFocusAnchor` property.
- `ACRActionOverflowRenderer` sets it to the button that opened the menu.
- The Cancel handler posts `UIAccessibilityLayoutChangedNotification` against that anchor.

Weak, so the target never keeps the button alive. Nil-checked, so a released anchor posts nothing rather than an unanchored notification.

Three files, +19/-1. No behaviour change for cards without an overflow menu.

## Evidence — and what it does not cover

VoiceOver element scan of the overflow scenario across three states, captured through the real `UIAccessibility` API (`XCUIElement`) on an iOS simulator:

| state | elements | hittable buttons |
|---|---|---|
| before open | 41 | 8 |
| menu open | 41 | **0** |
| after cancel | 41 | 8 |

Hittable buttons going `8 → 0 → 8` confirms the sheet really is presented and really is dismissed. The `after cancel` tree is byte-identical to `before open`, so this change introduces no regression to the card's accessibility tree.

**What this does not show.** The harness records `label / role / frame / isHittable / isEnabled / isSelected / value`. It does **not** record which element holds VoiceOver focus, and the action sheet is presented in a separate `UIWindow` that the scan does not traverse — note `Cancel` never appears in any of the three dumps.

So the focus restoration itself is argued from code inspection and from the existing `ACRShowCardTarget` precedent — it is **not** measured here. Calling that out plainly rather than implying a measurement that was not taken. Confirming it end to end needs VoiceOver driven on a physical device.

### Annotated overlays

Before opening the menu:

before

Menu open:

open

After cancel:

after dismiss

Raw per-element dumps:
- [before](https://hggzm.github.io/Teams-AdaptiveCards-Mobile/a11ymas_5536765_actionmode_cancel_before_elements.json)
- [open](https://hggzm.github.io/Teams-AdaptiveCards-Mobile/a11ymas_5536765_actionmode_cancel_after_elements.json)
- [after cancel](https://hggzm.github.io/Teams-AdaptiveCards-Mobile/a11ymas_5536765_actionmode_cancel_afterdismiss_elements.json)

## Scope

One behaviour: where VoiceOver focus goes when the overflow menu is cancelled. Cards without an overflow menu are unaffected, and the anchor is nil-safe.

Work item: **AB#5536765** (A11yMAS)

### DIFF
--- a/source/ios/AdaptiveCards/AdaptiveCards/AdaptiveCards/ACROverflowTarget.mm
+++ b/source/ios/AdaptiveCards/AdaptiveCards/AdaptiveCards/ACROverflowTarget.mm
@@ -242,9 +242,20 @@ NSString *const ACROverflowTargetIsRootLevelKey = @"isAtRootLevel";
         }
     }
 
+    // Dismissing the sheet has to say where focus goes. With a nil handler nothing posts
+    // an accessibility notification, so VoiceOver falls back to the platform default and
+    // the user loses their place instead of returning to the control they opened the menu
+    // from. ACRShowCardTarget already handles the equivalent case this way.
+    __weak __typeof(self) weakSelf = self;
     [_alert addAction:[UIAlertAction actionWithTitle:@"Cancel"
                                                style:UIAlertActionStyleCancel
-                                             handler:nil]];
+                                             handler:^(UIAlertAction *action) {
+                                                 __strong __typeof(weakSelf) strongSelf = weakSelf;
+                                                 UIView *anchor = strongSelf.accessibilityFocusAnchor;
+                                                 if (anchor) {
+                                                     UIAccessibilityPostNotification(UIAccessibilityLayoutChangedNotification, anchor);
+                                                 }
+                                             }]];
 }
 
 - (void)setInputs:(NSMutableArray *)inputs
--- a/source/ios/AdaptiveCards/AdaptiveCards/AdaptiveCards/ACROverflowTarget.mm
+++ b/source/ios/AdaptiveCards/AdaptiveCards/AdaptiveCards/ACROverflowTarget.mm
@@ -249,13 +249,17 @@ NSString *const ACROverflowTargetIsRootLevelKey = @"isAtRootLevel";
     __weak __typeof(self) weakSelf = self;
     [_alert addAction:[UIAlertAction actionWithTitle:@"Cancel"
                                                style:UIAlertActionStyleCancel
-                                             handler:^(UIAlertAction *action) {
-                                                 __strong __typeof(weakSelf) strongSelf = weakSelf;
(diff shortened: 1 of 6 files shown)


source/ios/AdaptiveCards/AdaptiveCards/AdaptiveCards/ACROverflowTarget.mm
```diff
@@ -242,9 +242,20 @@ - (void)createMenu:(ACOActionOverflow *)actionElement
...
 
+    // Dismissing the sheet has to say where focus goes. With a nil handler nothing posts
+    // an accessibility notification, so VoiceOver falls back to the platform default and
+    // the user loses their place instead of returning to the control they opened the menu
+    // from. ACRShowCardTarget already handles the equivalent case this way.
+    __weak __typeof(self) weakSelf = self;
     [_alert addAction:[UIAlertAction actionWithTitle:@"Cancel"
                                                style:UIAlertActionStyleCancel
-                                             handler:nil]];
+                                             handler:^(UIAlertAction *action) {
```
VikrantSingh01: Can we add a backing test for this change.

gauravkeshre (approved):

source/ios/AdaptiveCards/AdaptiveCards/AdaptiveCards/ACROverflowTarget.mm
hggzm: Added ACROverflowTargetTests.mm (7 tests). To avoid reaching into UIAlertAction's private handler ivar via KVC from a test, the restoration is extracted into a named method restoreAccessibilityFocusToAnchor, which the cancel handler now calls. Coverage: the sheet offers a cancel action, the anchor round-trips, the anchor is weak and does not retain the button, restoration is safe both with and without a live anchor, and the target still deallocates. One honest caveat: the tests assert the restoration path runs and is safe, not that VoiceOver received the notification - UIAccessibility notifications are delivered to the accessibility server and are not observable from a unit test with VoiceOver off. Happy to adjust the shape if you would rather keep the handler inline.

PR #738, opened 2026-08-27
[iOS][Fix] Luminosity issue fix

arocc:
# Description

Luminosity ratio of the placeholder text (Please Choose) is 1.674:1 which is less than 4.5:1.

**Repro Steps:**
1. Launch Adaptive Cards iOS application.
2. Home screen will appear.
3. Activate voice over
4. Navigate to "V1.5" control and select it.
5. Navigate to 'Scenarios' control and activate it.
6. Navigate to 'Restaurant Order' and activate it.
7. Perform right swipe and navigate to the 'Which side would you like?' combo box.
8. Open the tool and check the contrast ratio for the placeholder text and observe the issue.

# Sample Card

```json
{
	"type": "AdaptiveCard",
	"version": "1.5",
	"body": [
		{
			"type": "TextBlock",
			"text": "Malt & Vine Order Form",
			"size": "large",
			"wrap": true,
			"weight": "bolder",
			"style": "heading"
		},
		{
			"type": "Input.ChoiceSet",
			"id": "EntreeSelectVal",
			"label": "Which entree would you like?",
			"style": "filtered",
			"isRequired": true,
			"errorMessage": "This is a required input",
			"placeholder": "Please choose",
			"choices": [
				{
					"title": "Steak",
					"value": "1"
				},
				{
					"title": "Chicken",
					"value": "2"
				},
				{
					"title": "Tofu",
					"value": "3"
				}
			]
		}
	],
	"actions": [
		{
			"type": "Action.Submit",
			"title": "Place Order"
		}
	]
}
```

# How Verified

**Before**
"Please choose" text
image

Clear text field button
image

"Accent Tint" text
image

"Attention Tint" text
image

"Subtle Filled" and "Subtle Tint" text
image

**After**
"Please choose" text
Avalable Samole SONS

Clear text field button (passes WCAG requirement of at least 3:1 contrast for non-text graphical control)
image

"Accent Tint" text
image

"Attention Tint" text
image

"Subtle Filled" and "Subtle Tint" text
image

### DIFF
--- a/samples/HostConfig/microsoft-teams-dark.json
+++ b/samples/HostConfig/microsoft-teams-dark.json
@@ -370,7 +370,7 @@
 			"tint": {
 				"backgroundColor": "#2f2f4a",
 				"strokeColor": "#292929",
-				"textColor": "#7f85f5"
+				"textColor": "#a6a7dc"
 			}
 		},
 		"attention": {
@@ -382,7 +382,7 @@
 			"tint": {
 				"backgroundColor": "#250909",
 				"strokeColor": "#292929",
-				"textColor": "#e83a3a"
+				"textColor": "#f1707b"
 			}
 		},
 		"good": {
@@ -401,12 +401,12 @@
 			"filled": {
 				"backgroundColor": "#292929",
 				"strokeColor": "#292929",
-				"textColor": "#919191"
+				"textColor": "#adadad"
 			},
 			"tint": {
 				"backgroundColor": "#212121",
 				"strokeColor": "#292929",
-				"textColor": "#919191"
+				"textColor": "#adadad"
 			}
 		},
 		"subtle": {
--- a/samples/HostConfig/microsoft-teams-light.json
+++ b/samples/HostConfig/microsoft-teams-light.json
@@ -370,7 +370,7 @@
 			"tint": {
 				"backgroundColor": "#e8ebfa",
 				"strokeColor": "#e1e1e1",
-				"textColor": "#5b5fc7"
+				"textColor": "#444791"
 			}
 		},
 		"attention": {
@@ -382,7 +382,7 @@
 			"tint": {
 				"backgroundColor": "#f9d9d9",
 				"strokeColor": "#e1e1e1",
-				"textColor": "#d92c2c"
+				"textColor": "#a4262c"
 			}
 		},
 		"good": {
@@ -394,7 +394,7 @@
 			"tint": {
 				"backgroundColor": "#e7f2da",
 				"strokeColor": "#e1e1e1",
-				"textColor": "#0f7a0b"
+				"textColor": "#0b6a0b"
 			}
 		},
 		"informative": {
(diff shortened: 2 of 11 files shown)


rahulpinjani1 (approved):

hggzm (approved):
Approving — this addresses everything raised and then some. Thanks for the thorough pass.

I re-derived every changed host config token independently (WCAG 2.1 relative luminance, foreground against its own declared background):

| token | before | after | verdict |
|---|---|---|---|
| dark accent.tint | `#7f85f5` 4.043:1 | `#a6a7dc` **5.646:1** | pass |
| dark attention.tint | `#e83a3a` 4.544:1 | `#f1707b` **6.547:1** | pass |
| dark warning.filled | `#919191` 4.616:1 | `#adadad` **6.483:1** | pass |
| dark warning.tint | `#919191` 5.109:1 | `#adadad` **7.175:1** | pass |
| light accent.tint | `#5b5fc7` 4.530:1 | `#444791` **6.923:1** | pass |
| light attention.tint | `#d92c2c` 3.670:1 | `#a4262c` **5.517:1** | pass |
| light good.tint | `#0f7a0b` 4.755:1 | `#0b6a0b` **5.879:1** | pass |
| light warning.filled | `#6e6e6e` 4.514:1 | `#5f5f5f` **5.653:1** | pass |
| light warning.tint | `#6e6e6e` 4.514:1 | `#5f5f5f` **5.653:1** | pass |

**Nine improved, none regressed, all now clear 4.5:1.** Three were below AA before (`light attention.tint` at 3.670 and `dark accent.tint` at 4.043 being the worst), and the ones that already passed had almost no headroom — `light warning` at 4.514 was 0.014 above the line.

Three things I think are notably right here:

**Fixing it in the host config rather than at the call site.** These tokens are shared, so every element using accent/attention/warning tint benefits, not just the ChoiceSet. That is why this can close the other two items rather than just this one.

**`getTextBlockColor:...subtleOption:YES` instead of a constant.** The placeholder now follows the theme, so light and dark are both covered by construction and there is no hardcoded grey to keep in sync per appearance.

**The clear button.** Swapping `clearButtonMode` for a custom `UIButton` is the right call given UIKit does not let you tint the built-in one, and adding `accessibilityLabel = @"Clear text"` fixes a second problem — that control previously had no accessible name at all. `rightViewRectForBounds:` deferring to `clearButtonRectForBounds:` keeps the position identical, which I appreciated.

Also glad to see the tests assert `ACRContrastRatio(...) >= 4.5` in both appearances rather than equality against the constant. That will catch a regression that changes the value; an equality assertion would just need updating alongside it.

For the record: our earlier "not reproduced, 14.088:1" on this item was wrong — our tooling read the field's `textColor` rather than the placeholder, which is a separate property. Your 1.674:1 was right. We have since extended our accessibility harness to sample every foreground a user can actually see and to measure light and dark in the same run; it now independently reports **1.695:1** on these exact three ChoiceSets, which matches your figure. So this landed a real fix and improved our instrumentation.

PR #741, opened 2026-09-09
fix(ios): preserve Hours value when the field gains focus

gauravkeshre:
## Summary
Fixes the Hours numeric field losing its value when it gains focus.

## Root cause
`ACRNumericTextField` was setting `clearsOnBeginEditing = YES` during initialization, which caused the value to clear on begin editing.

## Fix
Changed `clearsOnBeginEditing` to `NO` so the existing Hours value is preserved when the field receives focus.

## Validation
- Added a regression test.
- Targeted test passed: `AdaptiveCardsTests/AdaptiveCardsTests/testNumericTextFieldPreservesValueWhenEditingBegins`

## Work item
https://domoreexp.visualstudio.com/MSTeams/_workitems/edit/5770572


hggzm (approved):
the fix looks good. the proxy reproduced the value loss on the baseline 3/3 times and passed with this change 3/3, including editing, refocus and submission. the accessible name stayed unchanged.

one non-blocking test suggestion: the new test checks the configuration but never begins editing. worth adding an actual focus/refocus transition so it covers the behavior its name describes. rest of the pr looks good to me.

P-Harika (approved):

